PR #9, opened 2015-03-07
Adjust timer for OS X

nerdcha:
Building the current master causes this error on OS X Yosemite:

    c++ -c -O3 -msse2 -fPIC -Wall -Wextra -Wno-unused-parameter -Wno-unknown-pragmas -pedantic   -o engine_mock.o src/engine_mock.cc
    In file included from src/engine_mock.cc:14:
    In file included from src/./allreduce_mock.h:14:
    src/../include/rabit/timer.h:18:30: error: use of undeclared identifier
          'CLOCK_REALTIME'
      utils::Check(clock_gettime(CLOCK_REALTIME, &ts) == 0, "failed to get time");
                             ^
    1 error generated.
    make: *** [engine_mock.o] Error 1

I've dropped in a fix, based on a [Stack Overflow answer](http://stackoverflow.com/questions/5167269/clock-gettime-alternative-in-mac-os-x). Not sure if there are any particular ways I should test it, but I can run the speed_test in rabit, and also build and use xgboost. (Thanks for xgboost, by the way; very nice!)

### DIFF
--- a/include/rabit/timer.h
+++ b/include/rabit/timer.h
@@ -8,14 +8,47 @@
 #include <time.h>
 #include "./utils.h"
 
+#ifdef __APPLE__
+#include <mach/mach_time.h>
+#define ORWL_NANO (+1.0E-9)
+#define ORWL_GIGA UINT64_C(1000000000)
+
+static double orwl_timebase = 0.0;
+static uint64_t orwl_timestart = 0;
+#endif  /* __APPLE__ */
+
 namespace rabit {
 namespace utils {
+    
+#ifdef __APPLE__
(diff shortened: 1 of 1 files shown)


tqchen:
Thanks for the update! It may not be good to put static member in header though. I will do a quick fix on this

tqchen:
I pushed the fix, xgboost should be fine now. you can test by run make xgboost.mock

nerdcha:
Thanks for that! Yes, rabit now builds correctly, and xgboost.mock builds as well if you copy this rabit into the xgboost subtree. (Plus guide-python/runall.sh executes ok.)

PR #13, opened 2015-05-20
use int32_t to define int and int64_t to define long. in VC long is 3…

hjk41:
in vc, long is just 32bit. I think it makes more sense to use int32_t to define kInt and use int64_t to define kLong

### DIFF
--- a/include/rabit/engine.h
+++ b/include/rabit/engine.h
@@ -183,7 +183,9 @@ enum DataType {
   kLong = 4,
   kULong = 5,
   kFloat = 6,
-  kDouble = 7
+  kDouble = 7,
+  kLongLong = 8,
+  kULongLong = 9
 };
 }  // namespace mpi
 /*!
--- a/include/rabit/rabit-inl.h
+++ b/include/rabit/rabit-inl.h
@@ -27,19 +27,19 @@ inline DataType GetType<unsigned char>(void) {
   return kUChar;
 }
 template<>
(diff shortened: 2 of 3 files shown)


tqchen:
This was used for redirect the calls to MPI. Maybe I was wrong on this. Shouldn't MPI_LONG be of same type as long in compiler? or not

hjk41:
I see. MPI specification requires that MPI_LONG = long in C. Then we should
define MPI_LONG_LONG, since size_t is defined as long long in VC

On Wed, May 20, 2015 at 10:14 AM, Tianqi Chen <notifications@github.com>
wrote:

> This was used for redirect the calls to MPI. Maybe I was wrong on this.
> Shouldn't MPI_LONG be of same type as long in compiler? or not
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/dmlc/rabit/pull/13#issuecomment-103727545>.
>



-- 
HONG Chuntao
System Research Group
Microsoft Research Asia

tqchen:
ic, can you update the PR for this?

hjk41:
sure

On Wed, May 20, 2015 at 10:27 AM, Tianqi Chen <notifications@github.com>
wrote:

> ic, can you update the PR for this?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/dmlc/rabit/pull/13#issuecomment-103729785>.
>



-- 
HONG Chuntao
System Research Group
Microsoft Research Asia

PR #28, opened 2016-07-27
Use getaddrinfo instead of gethostbyname for thread safety

randomjohnnyh:
See dmlc/xgboost#1394 for context.

I'm assuming that Rabit is intended to support running multiple Rabit threads. This is useful with xgboost4j-spark since Spark usually uses many cores per executor.

I'm surprised this issue hasn't shown up before.


tqchen:
Thanks for the PR!

tqchen:
please also bring a PR to update rabit version on the xgboost side

PR #38, opened 2016-11-23
allreduce_robust.cc: Allow num_global_replica to be 0

AbdealiJK:
In some cases, users may not want to have any global replica of
the data being broadcasted/all-reduced. In such cases, set the
result_buffer_round to -1 as a flag that this is not necessary
and check for it.

### DIFF
--- a/src/allreduce_robust.cc
+++ b/src/allreduce_robust.cc
@@ -33,7 +33,11 @@ AllreduceRobust::AllreduceRobust(void) {
 }
 void AllreduceRobust::Init(int argc, char* argv[]) {
   AllreduceBase::Init(argc, argv);
-  result_buffer_round = std::max(world_size / num_global_replica, 1);
+  if (num_global_replica == 0) {
+    result_buffer_round = -1;
+  } else {
+    result_buffer_round = std::max(world_size / num_global_replica, 1);
+  }
 }
 /*! \brief shutdown the engine */
 void AllreduceRobust::Shutdown(void) {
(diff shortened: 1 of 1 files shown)


tqchen:
please do a rebase so that travis can restarts checks, there a few lint errors that you might want to fix

AbdealiJK:
@tqchen Do you mean the guard for the header files ?
I can fix them if you like (I thought you might have prefered it in the way it is there right now. Hence left it as is)

src/allreduce_robust.cc
```diff
@@ -33,7 +33,11 @@ AllreduceRobust::AllreduceRobust(void) {
 }
 void AllreduceRobust::Init(int argc, char* argv[]) {
   AllreduceBase::Init(argc, argv);
-  result_buffer_round = std::max(world_size / num_global_replica, 1);
+  if ( num_global_replica == 0 ) {
```
tqchen: extra space in if

AbdealiJK:
@tqchen Wanted to confirm here - setting the num_global_replica to 0 shouldn't break anything else in the rest of the library right ? i.e. That is a valid config if I want no replicas of my data to be stored ?

tqchen:
I think it is valid, except that there is no fault tolerance now

AbdealiJK:
But there is still fault tolerance inside a single broadcast call and a single all reduce call. (As there is a loop inside Which checks and reinitiates the broadcast/all reduce) Which is what I want.

tqchen:
I see, that is correct

PR #45, opened 2017-06-06
Add build support for cygwin

sunnijie:
1. use `std::fopen` instead of `fopen64` under cygwin
2. turn off `DMLC_LOG_STACK_TRACE` under cygwin to fix error: "execinfo.h: No such file or directory"

### DIFF
--- a/include/dmlc/base.h
+++ b/include/dmlc/base.h
@@ -41,7 +41,7 @@
  * \brief Wheter to print stack trace for fatal error,
  * enabled on linux when using gcc.
  */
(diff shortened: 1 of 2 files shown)


sunnijie:
Not sure why Travis CI failed. I tried the code before my commits, still fails at the same point.

trivialfis:
I think this is handled in dmlc-core itself now.  And besides, I'm not keen on the idea of copying part of dmlc-core to here.  Should remove them eventually.

PR #51, opened 2017-12-05
Add option to fail gracefull on Rabit::Init failure

ebernhardson:
When running xgboost4j-spark the process that rabit
is initialized in has other things going on, such as
caches of data that is expensive to recompute. On
the rare occasion that rabit fails to connect to the
tracker rabit performs an exit(-1) which throws away
everything that was going on in the application.

It would be nice to somehow provide explicit failure
handling everywhere, but changing the external api
to that level would be quite a large breaking change.
This patch takes a very targeted approach changing
only the Init call to return a boolean indicating
success. This is disabled by default and must be
provided as part of the initialization parameters.

### DIFF
--- a/doc/guide.md
+++ b/doc/guide.md
@@ -25,7 +25,9 @@ using namespace rabit;
 const int N = 3;
 int main(int argc, char *argv[]) {
   int a[N];
-  rabit::Init(argc, argv);
+  if (!rabit::Init(argc, argv)) {
+    return -1;
+  }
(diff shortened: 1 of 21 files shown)


ebernhardson:
This doesn't work as is, more just throwing an idea up there.  Before i spend more time on this I wanted to see if this is something desirable, and if this is a reasonable direction or if something else would be more appropriate?

hcho3:

PR #57, opened 2018-04-20
Replace select() with poll() in SelectHandler

frenzykryger:
I use xgboost4j-spark to train several XGBoost instances sequentially one after another using 2 workers each with 4 threads. 
I found that after training 5-7 models process aborts with error described in https://github.com/dmlc/xgboost/issues/2736
This is due to SelectHandler using select() to watch for open sockets.

When launching multiple distributed XGBoosts within
one spark job it's pretty easy to run beyond FD_SETSIZE=1024
on linux, because fd number is always increases and it's not just sockets, but file descriptors too.

### DIFF
--- a/src/socket.h
+++ b/src/socket.h
@@ -20,17 +20,22 @@
 #include <arpa/inet.h>
 #include <netinet/in.h>
 #include <sys/socket.h>
-#include <sys/select.h>
 #include <sys/ioctl.h>
 #endif
 #include <string>
 #include <cstring>
+#include <vector>
+#include <unordered_map>
 #include "../include/rabit/internal/utils.h"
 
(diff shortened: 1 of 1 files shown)


nitinkak001:
I am getting this error while running XGBoost(0.72) through pyspark(2.3) shell. This happens when I run the job twice. First time it succeeds, second time the same job on the same data fails. I have tested it around 5 times now. Its only the second time it fails. Has there been any progress on this issue?

CodingCat:
close due to the merge of https://github.com/dmlc/rabit/pull/73

chenqin:
Thanks for @frenzykryger contribution! your diff has been very helpful.

PR #68, opened 2018-10-19
rabit support of large cluster

chenqin:
This is a cleanup and fix error based on #57 as well as minor tweak of tracker reconnect exponential back off retry. I were able to run on 1500 yarn containers instead of initial 350 with FD assertion failure.


chenqin:
close for now

PR #71, opened 2018-10-19
disable travis model_recover tests, fix doc generate failure

chenqin:


CodingCat:
@tqchen  what do you think? 

the failed model_recover test might also indicate the reason that why we never see rabit recover correctly in xgboost-spark so that xgboost-spark has its own checkpoint mechanism

tqchen:
I am going to merge this in. The recover test is failing because some of the script semantics changed

PR #73, opened 2018-10-21
support larger cluster

chenqin:
We run some experiments with xgboost4j-spark, found rabbit will throw FD assertion failure at around 350 executers. This is largely due to what #57 described and seem only need to change from select to poll would avoid such limit from OS side. In our experiment, we were able to scale to 1.5k executors on 12 Billions dataset after some tweaks here and there.


CodingCat:
thanks for the contribution, merge this in, credits also to @frenzykryger

PR #74, opened 2018-10-26
workaround macosx java test race condition

chenqin:
https://github.com/dmlc/xgboost/pull/3818


src/allreduce_robust.cc
```diff
@@ -50,6 +50,9 @@ void AllreduceRobust::Shutdown(void) {
   // execute check ack step, load happens here
   utils::Assert(RecoverExec(NULL, 0, ActionSummary::kCheckAck, ActionSummary::kSpecialOp),
                 "Shutdown: check ack must return true");
+#ifdef __APPLE__
+  sleep(2);
```
CodingCat: comments?

chenqin: fixed

PR #77, opened 2018-12-30
Try fixing mingw build error when using CMake.

trivialfis:
* Check __MINGW32__ .

I tried to use `CMakeLists.txt` from `rabit` in `XGBoost` by calling `add_subdirectory`, but the build failed on mingw.  I'm not sure this will fix it since I don't have mingw currently for testing.  Please close this if there's a better solution. :)

Here is the build log:
https://ci.appveyor.com/project/tqchen/xgboost/builds/21295997/job/lip9i83alwamavv5


chenqin:
@tqchen , BTW, I am interested to review changes and fixing out of dated issues in rabit moving forward.

PR #80, opened 2019-02-17
Fix ssize_t definition.

trivialfis:


hcho3:
@trivialfis Do we need this for XGBoost?

trivialfis:
@hcho3 Yes.  I need to upgrade rabit for xgboost.  Will address the linter.

trivialfis:
@hcho3 Any chance that you can add a appvayor test for rabit?  I'm currently relying on to make sure this PR is done right: https://github.com/dmlc/xgboost/pull/4159

trivialfis:
And by the way, if there's need to address the linter in `dmlc-core`, we should write a script to do so ...

PR #83, opened 2019-03-15
[rabit harden] add engine_mpi build

chenqin:
add travis task mpi-build against engine_mpi and speed_test.mpi 
@trivialfis


trivialfis:
@chenqin I think MPI is no longer maintaining a c++ interface which Rabit depends on.  Do we want to build a wrapper inside Rabit?

PR #86, opened 2019-03-19
[rabit harden] replace hardcopy dmlc-core headers with submodule links

chenqin:
dmlc-core headers was unsync with rabit dmlc headers https://github.com/dmlc/rabit/pull/78 , this cause rabit mpi build errors. 
```
xxx:test cq$ make
cd ..;make;cd -
make[1]: Nothing to be done for `all'.
/Users/cq/xgboost/rabit/test
cd ..;make mpi;cd -
mpicxx -c -O3 -Wall -Wextra -Wno-unused-parameter -Wno-unknown-pragmas -std=c++11 -msse2 -fPIC -o engine_mpi.o src/engine_mpi.cc
In file included from src/engine_mpi.cc:14:
In file included from src/../include/rabit/internal/engine.h:10:
In file included from src/../include/rabit/internal/../serializable.h:12:
src/../include/rabit/internal/../../dmlc/io.h:15:10: fatal error: 'dmlc/base.h' file not found
#include "dmlc/base.h"
         ^~~~~~~~~~~~~
1 error generated.
make[1]: *** [engine_mpi.o] Error 1
```

https://github.com/chenqin/rabit/commit/ed06e0c6af9444ddf570291bc383fc701b9286b4#diff-bbc45ff15fdaa4016cdcf3cc7447bec5L12 was introduced to address rabit test failures. 

At same time, a bug was introduced which cause XGB build with latest rabit master failure, detail can be found https://github.com/dmlc/xgboost/issues/4274

This pr backport latest dmlc-core header files used by rabit, to keep interface definition consistent while building xgb with dmlc and rabit headers. This is important as xgb will includes both rabit and dmlc-core as part of headers definition yet link to dmlc-core lib 

Here is verification chenqin/xgboost ci build point to chenqin/rabit 
https://travis-ci.org/chenqin/xgboost/builds/508611790

PR #90, opened 2019-04-09
[rabit harden] include osx in tests, address time_wait on port assignment, enable all tests

chenqin:
follow up clean up dmlc-core header file copy pr https://github.com/chenqin/rabit/commit/ecd4bf7aaee477cea6ac14aa52f8276192f085f6.
- adding osx as OS on all ci tasks
- split mpi test from socket ones
- address tcp socket time_wait issue
- enable all tests including python
- address tracker deadlock in python2 via upgrade tracker to python3 https://github.com/dmlc/dmlc-core/pull/525
more detail of deadlock rootcause can be found https://github.com/dmlc/dmlc-core/pull/524


.travis.yml
```diff
@@ -10,12 +15,27 @@ env:
     - TASK=build
     - TASK=mpi-build
     - TASK=cmake-build
-    - TASK=test CXX=g++-4.8
+    - TASK=test CXX=g++
+    - TASK=xgb-java-tests
+
+matrix:
+  exclude:
```
CodingCat: why these are excluded?

.travis.yml
```diff
@@ -31,30 +51,44 @@ addons:
       - openssh-client
       - openssh-server
       - libopenmpi-dev
+      - clang
+      - clang-tidy-5.0
+      - cmake-data
+      - graphviz
+  homebrew:
+    packages:
+      - gcc@7
```
CodingCat: this is a very high version of gcc requirement @hcho3 are we just requiring gcc 5 in xgb?

.travis.yml
```diff
@@ -31,30 +51,44 @@ addons:
...
+      - graphviz
```
CodingCat: why?

.travis.yml
```diff
@@ -10,12 +15,27 @@ env:
...
+    - os: osx
+      env: TASK=doc
+    - os: osx
+      env: TASK=mpi-build
 
 # dependent apt packages
-dist: xenial
 addons:
   apt:
+    sources:
```
CodingCat: can you specify the reason for adding each of dependencies here?

.travis.yml
chenqin: ported from xgb master https://github.com/dmlc/xgboost/blob/master/.travis.yml#L71

.travis.yml
chenqin: https://github.com/chenqin/rabit/blob/master/.travis.yml#L22
exclude osx lint and doc as well as mpi, for first three since we already do in linux so it's matter of save build time. For mpi, I don't see any use case running mpi in osx, we do have linux tests running btw.

chenqin:
> have we fixed the test mentioned in [#86 (comment)](https://github.com/dmlc/rabit/pull/86#issuecomment-475481175)?

That should be another pr after we updated xgboost master. The rationale is previous and this pr haven't change how rabit functions other than reshuffle parameters and delete redundant code.

I already got some idea how to fix flaky test by introducing extra check in allreduce_robust before reset links infinitely. But yeah, I think that should be seperate thing after we have good baseline https://github.com/dmlc/xgboost/pull/4352

.travis.yml
chenqin: same as above, graphviz seems redudant at this point. It's within my plan to output allreduce toplogy graph as some point in next couple of prs.

scripts/travis_script.sh
```diff
@@ -30,4 +30,17 @@ if [ ${TASK} == "cmake-build" ]; then
     cd build
     cmake ..
     make all || exit -1
-fi
\ No newline at end of file
+fi
+
+if [ ${TASK} == "xgb-java-tests" ]; then
```
CodingCat: are we really going to do this?

it sounds strange to me....Spark relies on netty but netty never put Spark as a integration test, 

@hcho3  what do you think

hcho3: @CodingCat This is something I've been thinking about, about the need for integration tests for dmlc-core / rabit against upstream projects. The question is where is the correct place to perform such tests. Should Rabit host integration tests for XGBoost, or should XGBoost do it? Or should we create a separate GitHub repo to do inter-project integration tests?

chenqin: removed xgb java tests from rabit. So integration test fall back to where we started and only runs when we try to upgrade rabit in xgboost. I incline to keep this process as simple as possible with cost of manual work on upgrade rabit from time to time.

CodingCat: I think XGBoost should do an integration test covering all involved rabit functionalities

Rabit (though XGB is the only significant consumer for now) should go as it is, and when we upgrade Rabit in XGB, the integration test there should be a guard to prevent regressions leading to the failure of XGB (not Rabit)

There are several cases we should cover 

1. XGB has breaking changes blocking the upgrading to the latest version of Rabit, integration test should not be able to run, so that we can fix XGB

2. Rabit has some bug not triggered by XGB, XGB just upgrades as long as we are sure integration test covers all involved functionalities 

3. Rabit has a bug triggered by XGB, integration test should detect that and we should fix in Rabit

I think it is the common practice in other projects, 

1. Hadoop 3 is very different with Hadoop 2, so Spark has to bring a lot of changes to upgrade to that

2. Parquet/Orc has several bugs all the way but not triggered by Spark, Spark just upgrade as usual

3. Orc has a bug, which is discovered by some integration test in some company, makes Spark cannot upgrade to the latest version, then Spark postpone the upgrading and Orc folks in Spark community went back to fix and added a test to guard

chenqin:
related test failure, here are some investigation. 
1) for somereason, trybind actually allows two process bind to same port. I fixed this with another constructor like
`
SockAddr(int port) {
    addr.sin_family = AF_INET; 
    addr.sin_addr.s_addr = INADDR_ANY; 
    addr.sin_port = htons( port ); 
  }
`
2) when recovered rabit worker couting number of connection rank 9 as shown below never get connection accepted **num_accept** hence stuck forever.

```
cq@cq-MS-7B84:~/xgboost/rabit/test$ make -f test.mk 
../dmlc-core/tracker/dmlc-submit --cluster local --num-workers=10 --local-num-attempt=10 model_recover 10000 mock=0,0,1,0 mock=1,1,1,0
2019-04-10 15:17:51,502 INFO start listen on 10.0.1.25:9091
[0]ReConnectLinks cmd start port 9010 range (9010 - 10010)
[1]ReConnectLinks cmd start port 9011 range (9010 - 10010)
[0]sock_listen.Accept cmd start port # 9010 0 num_accept 2
[2]ReConnectLinks cmd start port 9012 range (9010 - 10010)
[1]sock_listen.Accept cmd start port # 9011 0 num_accept 2
[3]ReConnectLinks cmd start port 9013 range (9010 - 10010)
[2]sock_listen.Accept cmd start port # 9012 0 num_accept 2
[4]ReConnectLinks cmd start port 9014 range (9010 - 10010)
[2]sock_listen.Accept cmd start port # 9012 1 num_accept 2
[3]sock_listen.Accept cmd start port # 9013 0 num_accept 1
[5]ReConnectLinks cmd start port 9015 range (9010 - 10010)
[4]sock_listen.Accept cmd start port # 9014 0 num_accept 1
[6]ReConnectLinks cmd start port 9016 range (9010 - 10010)
[1]sock_listen.Accept cmd start port # 9011 1 num_accept 2
[5]sock_listen.Accept cmd start port # 9015 0 num_accept 1
[7]ReConnectLinks cmd start port 9017 range (9010 - 10010)
[6]sock_listen.Accept cmd start port # 9016 0 num_accept 1
[8]ReConnectLinks cmd start port 9018 range (9010 - 10010)
[7]sock_listen.Accept cmd start port # 9017 0 num_accept 2
[9]ReConnectLinks cmd start port 9019 range (9010 - 10010)
[0]sock_listen.Accept cmd start port # 9010 1 num_accept 2
[8]sock_listen.Accept cmd start port # 9018 0 num_accept 1
[7]sock_listen.Accept cmd start port # 9017 1 num_accept 2
[0] reload-trail=0, init iter=0
[1] reload-trail=0, init iter=0
[9] reload-trail=0, init iter=0
[7] reload-trail=0, init iter=0
[6] reload-trail=0, init iter=0
[2] reload-trail=0, init iter=0
[4] reload-trail=0, init iter=0
[5] reload-trail=0, init iter=0
2019-04-10 15:17:52,491 INFO @tracker All of 10 nodes getting started
[8] reload-trail=0, init iter=0
[3] reload-trail=0, init iter=0
[0] !!!TestMax pass, iter=0
[3] !!!TestMax pass, iter=0
[6] !!!TestMax pass, iter=0
[0]@@@Hit Mock Error:Broadcast
[4] !!!TestMax pass, iter=0
[5] !!!TestMax pass, iter=0
[2] !!!TestMax pass, iter=0
[8] !!!TestMax pass, iter=0
[9] !!!TestMax pass, iter=0
[7] !!!TestMax pass, iter=0
[1] !!!TestMax pass, iter=0
[9]ReConnectLinks cmd recover port 9014 range (9010 - 10010)
[8]ReConnectLinks cmd recover port 9015 range (9010 - 10010)
[9]sock_listen.Accept cmd recover port # 9014 0 num_accept 3
[7]ReConnectLinks cmd recover port 9016 range (9010 - 10010)
[8]sock_listen.Accept cmd recover port # 9015 0 num_accept 1
[9]sock_listen.Accept cmd recover port # 9014 1 num_accept 3
[1]ReConnectLinks cmd recover port 9017 range (9010 - 10010)
[3]ReConnectLinks cmd recover port 9018 range (9010 - 10010)
[2]ReConnectLinks cmd recover port 9019 range (9010 - 10010)
[1]sock_listen.Accept cmd recover port # 9017 0 num_accept 3
[3]sock_listen.Accept cmd recover port # 9018 0 num_accept 2
[6]ReConnectLinks cmd recover port 9020 range (9010 - 10010)
[1]sock_listen.Accept cmd recover port # 9017 1 num_accept 3
[7]sock_listen.Accept cmd recover port # 9016 0 num_accept 1
[5]ReConnectLinks cmd recover port 9021 range (9010 - 10010)
[6]sock_listen.Accept cmd recover port # 9020 0 num_accept 1
[4]ReConnectLinks cmd recover port 9022 range (9010 - 10010)
[2]sock_listen.Accept cmd recover port # 9019 0 num_accept 1
[3]sock_listen.Accept cmd recover port # 9018 1 num_accept 2
[5]sock_listen.Accept cmd recover port # 9021 0 num_accept 1

```

chenqin:
> have we fixed the test mentioned in [#86 (comment)](https://github.com/dmlc/rabit/pull/86#issuecomment-475481175)?

Done

chenqin:
osx build error due to unable to download package. should be fine if rerun. local build here.
https://travis-ci.org/chenqin/rabit/builds/518971901

trivialfis:
@chenqin thanks, restated the test.

.travis.yml
CodingCat: this is concerning, we have seen enough cases about different network behaviors in Mac/Linux

As a developer, test behaviors in mac os before deploying to linux production servers is a common behavior

I would suggest we pursue the consistency of tests in different OS with the best effort

chenqin:
all tests green  @CodingCat @trivialfis

CodingCat:
check my comments about mpi in osx

CodingCat:
hmmmm.....MPI in Mac doesn't compile....we missed some dependency?

and test itself is still flaky, I restarted linux one but mac one looks like stuck there (I didn't restart for your further investigation)

chenqin:
> hmmmm.....MPI in Mac doesn't compile....we missed some dependency?
Seems brew install can't do with cxx binding, needs to recompile from source everytime. Hopefully it can be done within 10mins before travis timeout. found this blog, going to try this path
https://d-meiser.github.io/2016/01/10/mpi-travis.html

> and test itself is still flaky, I restarted linux one (thought it was a Travis issue) but found mac one stuck there later (I didn't restart for your further investigation)

Will take a look, this time is mac

chenqin:
After some investigation, looks like the stuck is caused by tracker subprocess.call not return exit code despite c executable exit(-2).

chenqin:
`if os.name == 'nt':
            ret = subprocess.call(cmdline, shell=True, env=env)
        else:
            ret = subprocess.call(cmdline, shell=True, executable='bash', env=env)
            print ret
        if ret == 0:
            logging.debug('Thread %d exit with 0', taskid)
            return`

`
void Verify(const MockKey &key, const char *name) {
    if (mock_map.count(key) != 0) {
      num_trial += 1;
      fprintf(stderr, "[%d]@@@Hit Mock Error:%s\n", rank, name);
      memset((char *)0x0, 1, 100); abort(); exit(-2); //all tried
    }
`
expected failure follows tracker print exit code
`[1]@@@Hit Mock Error:Broadcast
[7] !!!TestMax pass, iter=1
[9] !!!TestMax pass, iter=1
-11
['./model_recover', '10000', 'mock=0,0,1,0', 'mock=1,1,1,0']
`

actually see no print exit code and stuck (one of worker did nothing basically)
`
../dmlc-core/tracker/dmlc-submit --cluster local --num-workers=10 --local-num-attempt=20 model_recover 10000 mock=0,0,1,0 mock=1,1,1,0 mock=0,1,1,0 mock=4,1,1,0 mock=9,1,1,0
2019-04-15 18:53:14,652 INFO start listen on 10.0.1.25:9091
[0] reload-trail=0, init iter=0
[5] reload-trail=0, init iter=0
[9] reload-trail=0, init iter=0
[8] reload-trail=0, init iter=0
[1] reload-trail=0, init iter=0
2019-04-15 18:53:15,632 INFO @tracker All of 10 nodes getting started
[6] reload-trail=0, init iter=0
[3] reload-trail=0, init iter=0
[4] reload-trail=0, init iter=0
[7] reload-trail=0, init iter=0
[2] reload-trail=0, init iter=0
[0] !!!TestMax pass, iter=0
[6] !!!TestMax pass, iter=0
[0]@@@Hit Mock Error:Broadcast
[2] !!!TestMax pass, iter=0
[8] !!!TestMax pass, iter=0
[5] !!!TestMax pass, iter=0
[4] !!!TestMax pass, iter=0
[7] !!!TestMax pass, iter=0
[9] !!!TestMax pass, iter=0
[1] !!!TestMax pass, iter=0
[3] !!!TestMax pass, iter=0
`

chenqin:
https://github.com/dmlc/dmlc-core/pull/524

CodingCat:
isn't the test still hanging?

chenqin:
We need to fix dmlc-core first.

CodingCat:
> We need to fix dmlc-core first.

check https://github.com/dmlc/dmlc-core/pull/525#issuecomment-484567068

CodingCat:
looks like we have very consistent behavior in linux with pip3

.travis.yml
```diff
@@ -4,7 +4,7 @@ os:
   - linux
   - osx
 
-osx_image: xcode10.2
+osx_image: xcode9.3
```
CodingCat: what' wrong with xcode10.2 actually?

scripts/mpi.sh
```diff
@@ -0,0 +1,27 @@
+#!/usr/bin/env bash
```
CodingCat: maybe rename to mpi_build.sh

test/test.mk
```diff
@@ -1,29 +1,29 @@
 # this is a makefile used to show testcases of rabit
 .PHONY: all
```
CodingCat: are we ok to re-enable pylocal_recover_10_10k now?

chenqin: I am considering refactor rabit python wrapper as some point in near future and consolidate some duplicated code in rabit and xgboost code base.

The observation is we got similar code in xgboost/python package as well as rabit itself. If we want to verify rabit python wrapper can work as we expected, we can utilize xgboost python tests.

scripts/mpi.sh
chenqin: fixed.

.travis.yml
chenqin: reverted.

test/test.mk
chenqin: done

chenqin:
friendly ping

trivialfis:
Merge.  I hope you don't mind me changing the commit message. ;-)  

I'm more worry about dependency on `dmlc-core`.  The other day I tried to use C++ API from mxnet, it turns out there's no way to install it from build system with so many duplicated dmlc-core as git submodules.  If the common directory is not common enough, we might just remove it completely, or try to use a proper package manager instead of git submodule.

PR #91, opened 2019-04-24
[one dmlc-core] by default build with dmlc-core from xgboost

chenqin:
- remove dmlc-core gitsubmodule from rabit
- use travis cmd & RABIT_BUILD_DMLC = 1 to build rabit with dmlc-core included

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -4,6 +4,7 @@ project(rabit VERSION 0.2.0)
 
 option(RABIT_BUILD_TESTS "Build rabit tests" OFF)
 option(RABIT_BUILD_MPI "Build MPI" OFF)
+option(RABIT_BUILD_DMLC "Include DMLC_CORE in build" ON)
 
 add_library(rabit src/allreduce_base.cc src/allreduce_robust.cc src/engine.cc src/c_api.cc)
 add_library(rabit_base src/allreduce_base.cc src/engine_base.cc src/c_api.cc)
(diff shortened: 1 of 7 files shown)


chenqin:
@trivialfis @CodingCat

CMakeLists.txt
```diff
@@ -26,13 +27,23 @@ if(RABIT_BUILD_TESTS)
...
+      "$<BUILD_INTERFACE:${CMAKE_CURRENT_LIST_DIR}/include>"
+      "$<BUILD_INTERFACE:${CMAKE_CURRENT_LIST_DIR}/dmlc-core/include>"
+      )
+  endforeach()
+else()
+  foreach(lib ${rabit_libs})
+    #include "./internal/utils.h"
+    target_include_directories(${lib} PUBLIC
+      "$<BUILD_INTERFACE:${CMAKE_CURRENT_LIST_DIR}/include>"
+      "$<BUILD_INTERFACE:${CMAKE_CURRENT_LIST_DIR}/../dmlc_core/include>"
```
CodingCat: may leave comments here about we assume rabit and dmlc-core are both put in the root path of the project which depends on it?

CodingCat: or we have some way to detect it

CodingCat: nvm, I think they only need to be in the same level of dir which is very likely to be the case

trivialfis:
@chenqin Thanks!  Have you tried it with XGBoost?

chenqin:
link to chenqin/rabit master, pass travis-ci
commit
https://github.com/chenqin/xgboost/commit/c3b8af10ed83e6bc53094c896542ba14268c835f
travis
https://travis-ci.org/chenqin/xgboost/builds/524781844

trivialfis:
Thanks!

include/rabit/serializable.h
```diff
@@ -9,7 +9,10 @@
 #include <vector>
 #include <string>
 #include "./internal/utils.h"
+
+#ifndef DMLC_IO_H_
 #include "../../dmlc-core/include/dmlc/io.h"
```
codingforfun: This is a no go. If you do this you can just put the source of dmlc and rabit together in a single project. 
Correct way to do this would be to just include `dmlc/io.h` and to let cmake correctly determine include dirs.

CMakeLists.txt
codingforfun: This is a really ugly hack. Don’t assume something outside of your project. This is what cmake is for. Set the correct include dir on the target using cmake.

include/rabit/serializable.h
chenqin: https://github.com/dmlc/rabit/pull/106/files

CMakeLists.txt
chenqin: https://github.com/dmlc/rabit/pull/106/files

PR #92, opened 2019-04-26
allow multi call on init

CodingCat:
@hcho3 as part of fix

we need to make init() idempotent , the reason is that in XGBoost-Spark , users may partially evaluate a DF/RDD through APIs like show(), and then evaluate fully for prediction/training

As a result, Rabit.Shutdown is not called in the partial evaluation as we only call it when we exhaust a complete partition (and it's hard to call it in the case of partial evaluation since we have no pre-knowledge about how many elements are evaluated). 

Finally, we call init() again in the case of full evaluation and the guard in init() function would crash the process


CodingCat:
I will submit XGBoost part PR after this is approved

PR #95, opened 2019-06-13
Allow not stop process on error

CodingCat:
in some distributed environment like Spark, multiple workers co-locate in the same process. For anyone of them to fail, it should not stop the whole process and take down the other co-locating workers

this PR essentially makes this behavior as an option instead of the only one in rabit

@hcho3 @trivialfis @RAMitchell @chenqin

### DIFF
--- a/include/rabit/internal/utils.h
+++ b/include/rabit/internal/utils.h
@@ -62,7 +62,7 @@ namespace utils {
 const int kPrintBuffer = 1 << 12;
 
 /*! \brief we may want to keep the process alive when there are multiple workers co-locate in the same process */
-bool STOP_PROCESS_ON_ERROR = true;
+extern bool STOP_PROCESS_ON_ERROR;
 
 #ifndef RABIT_CUSTOMIZE_MSG_
 /*!
--- a/include/rabit/internal/utils.h
+++ b/include/rabit/internal/utils.h
@@ -61,6 +61,9 @@ namespace utils {
 /*! \brief error message buffer length */
 const int kPrintBuffer = 1 << 12;
 
+/*! \brief we may want to keep the process alive when there are multiple workers co-locate in the same process */
+bool STOP_PROCESS_ON_ERROR = true;
+
 #ifndef RABIT_CUSTOMIZE_MSG_
 /*!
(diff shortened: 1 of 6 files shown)


RAMitchell:
Looks ok, global variables are evil but there's not another easy way to do this.

I didn't know that Spark could have multiple workers in a single process. Does this depend on some user configuration or can this happen as default behaviour? I am asking because in designing the GPU solution it seems much safer to have one process per GPU. Otherwise wouldn't it be possible for workers to modify global state such as omp_num_threads and have an effect on the other workers in the same process? cc @rongou

rongou:
If we statically allocate a GPU for each Spark executor and only launch a single task on each executor, then yes one process per GPU holds. But if we are allocating GPUs dynamically then I'm not sure. @tgravescs is working on GPU scheduling for Spark and might have a better answer.

CodingCat:
In CPU, you can specify how many cores you assigned to each executor of spark, and then you decide how many cores each task(worker) will take...if you make them equal, it will become one process one worker 

I believe there would be some way to constraint it with GPU resources

RAMitchell:
@CodingCat Hypothetically, would it be possible to enforce that xgboost on spark always uses separate processes for workers? We don't really have a lot of tests for running multiple xgboosts in a single process and a lot of our code is not thread safe.

CodingCat:
> @CodingCat Hypothetically, would it be possible to enforce that xgboost on spark always uses separate processes for workers? We don't really have a lot of tests for running multiple xgboosts in a single process and a lot of our code is not thread safe.

well, it's possible, but I would prefer to making it as an option to the user where they just set `nthread == spark.task.cpus', and to set it in our side, we can just check the parameter in xgboost spark code 

but limit one worker one process would increase the cost of memory in spark cluster per my test. Furthermore, a very practical reason is that all unit test in Spark side would not be runnable as in local mode, we have only one process (referring as Spark driver) and all tasks are running in this "Driver"

chenqin:
LGTM, i think we might just add this to dmlc_tracker/local.py and minic launch worker with threads instead of subprocess so we can integration test this.

CodingCat:
close for now to separate content into multiple PRs

PR #96, opened 2019-06-25
return values in Init and Finalize

CodingCat:
currently init and finalize methods in different engines are all void 

it creates a bunch of problems for cross-language calling as it always core dump in native layer for any error

this PR is to make functions ranging from engine::Init to underlying allreduce_*: Init return true or false

@RAMitchell @chenqin @trivialfis @hcho3

### DIFF
--- a/src/allreduce_base.cc
+++ b/src/allreduce_base.cc
@@ -57,7 +57,7 @@ AllreduceBase::AllreduceBase(void) {
 }
 
 // initialization function
-void AllreduceBase::Init(int argc, char* argv[]) {
+bool AllreduceBase::Init(int argc, char* argv[]) {
   // setup from enviroment variables
   // handler to get variables from env
   for (size_t i = 0; i < env_vars.size(); ++i) {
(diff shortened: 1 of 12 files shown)


src/allreduce_base.cc
```diff
@@ -116,24 +116,30 @@ void AllreduceBase::Init(int argc, char* argv[]) {
...
+    if (tracker_uri == "NULL") return true;
+    // notify tracker rank i have shutdown
+    utils::TCPSocket tracker = this->ConnectTracker();
+    tracker.SendStr(std::string("shutdown"));
+    tracker.Close();
+    // close listening sockets
+    sock_listen.Close();
+    utils::TCPSocket::Finalize();
+    return true;
+  } catch (std::exception e) {
```
trivialfis: Catch by reference (const reference).  Exception could be polymorphic class.

src/allreduce_base.cc
```diff
@@ -237,167 +243,179 @@ utils::TCPSocket AllreduceBase::ConnectTracker(void) const {
...
+           "ReConnectLink failure 14");
+    // close connection to tracker
+    tracker.Close();
+    // listen to incoming links
+    for (int i = 0; i < num_accept; ++i) {
+      LinkRecord r;
+      r.sock = sock_listen.Accept();
       Assert(r.sock.SendAll(&rank, sizeof(rank)) == sizeof(rank),
-             "ReConnectLink failure 12");
+             "ReConnectLink failure 15");
```
trivialfis: What are these numbers ...

PR #97, opened 2019-06-25
allow not stop process in error

CodingCat:
part  of the original https://github.com/dmlc/rabit/pull/95 which has involved https://github.com/dmlc/rabit/pull/96 as well 

in some distributed environment like Spark, multiple workers co-locate in the same process. For anyone of them to fail, it should not stop the whole process and take down the other co-locating workers

this PR essentially makes this behavior as an option instead of the only one in rabit


CodingCat:
merge this, as it is already got reviewed

PR #98, opened 2019-07-02
allreduce/broadcast cache

chenqin:
The goal of this pr is to implement immutable cache in rabit to help failed worker recover onetime allreduces in bootstrap time.  examples of such are
- distributed xgboost load dataset and sync number of columns before loadcheckpoint.
- init histgram in fast hist algorithm only runs before completion of first iteration checkpoint

It's specifically designed to help recovered node catch up with rest of nodes with minimal overhead in non recovery mode. 
- setcache only writes locally after allreduce/broadcast complete
- getcache send/recv two integer in one allreduce call and decide if all nodes are running in sync
- when nodes are in sync, skip reading cache, only recovery nodes in bootstrap do one time catch rebuild from nearest nodes

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,10 +1,10 @@
-cmake_minimum_required(VERSION 3.0)
+cmake_minimum_required(VERSION 3.3)
 
-project(rabit VERSION 0.2.0)
+project(rabit VERSION 0.2.1)
 
 option(RABIT_BUILD_TESTS "Build rabit tests" OFF)
 option(RABIT_BUILD_MPI "Build MPI" OFF)
-option(RABIT_BUILD_DMLC "Include DMLC_CORE in build" ON)
+option(RABIT_BUILD_DMLC "Include DMLC_CORE in build" OFF)
 
 add_library(rabit src/allreduce_base.cc src/allreduce_robust.cc src/engine.cc src/c_api.cc)
 add_library(rabit_base src/allreduce_base.cc src/engine_base.cc src/c_api.cc)
--- a/Makefile
+++ b/Makefile
@@ -9,7 +9,7 @@ else
 endif
 
 export WARNFLAGS= -Wall -Wextra -Wno-unused-parameter -Wno-unknown-pragmas -std=c++11
-export CFLAGS = -O3 -g $(WARNFLAGS) -I $(DMLC)/include -I include/
+export CFLAGS = -O3 $(WARNFLAGS) -I $(DMLC)/include -I include/
 export LDFLAGS =-Llib
 
 #download mpi
--- a/Makefile
+++ b/Makefile
@@ -9,7 +9,7 @@ else
 endif
 
 export WARNFLAGS= -Wall -Wextra -Wno-unused-parameter -Wno-unknown-pragmas -std=c++11
-export CFLAGS = -O3 $(WARNFLAGS) -I $(DMLC)/include -I include/
+export CFLAGS = -O3 -g $(WARNFLAGS) -I $(DMLC)/include -I include/
 export LDFLAGS =-Llib
 
 #download mpi
--- a/include/rabit/internal/rabit-inl.h
+++ b/include/rabit/internal/rabit-inl.h
@@ -156,9 +156,10 @@ inline void Broadcast(std::string *sendrecv_data, int root) {
 template<typename OP, typename DType>
 inline void Allreduce(DType *sendrecvbuf, size_t count,
                       void (*prepare_fun)(void *arg),
-                      void *prepare_arg) {
+                      void *prepare_arg,
+                      int* cache_seq) {
   engine::Allreduce_(sendrecvbuf, sizeof(DType), count, op::Reducer<OP, DType>,
-                     engine::mpi::GetType<DType>(), OP::kType, prepare_fun, prepare_arg);
+                     engine::mpi::GetType<DType>(), OP::kType, prepare_fun, prepare_arg, cache_seq);
 }
 
 // C++11 support for lambda prepare function
@@ -167,9 +168,9 @@ inline void InvokeLambda_(void *fun) {
   (*static_cast<std::function<void()>*>(fun))();
 }
 template<typename OP, typename DType>
-inline void Allreduce(DType *sendrecvbuf, size_t count, std::function<void()> prepare_fun) {
+inline void Allreduce(DType *sendrecvbuf, size_t count, std::function<void()> prepare_fun, int* cache_seq) {
   engine::Allreduce_(sendrecvbuf, sizeof(DType), count, op::Reducer<OP, DType>,
-                     engine::mpi::GetType<DType>(), OP::kType, InvokeLambda_, &prepare_fun);
+                     engine::mpi::GetType<DType>(), OP::kType, InvokeLambda_, &prepare_fun, cache_seq);
 }
 #endif  // C++11
 
@@ -178,12 +179,12 @@ inline void TrackerPrint(const std::string &msg) {
   engine::GetEngine()->TrackerPrint(msg);
 }
 
-inline void TrackerSetConfig(const std::string &key, const int &value) {
-  engine::GetEngine()->TrackerSetConfig(key, value);
+inline void TrackerSetCacheIndex(const std::string &key, const int &value) {
+  engine::GetEngine()->TrackerSetCacheIndex(key, value);
 }
 
-inline void TrackerGetConfig(const std::string &key, int &value) {
-  engine::GetEngine()->TrackerGetConfig(key, value);
+inline void TrackerGetCacheIndex(const std::string &key, int &value) {
+  engine::GetEngine()->TrackerGetCacheIndex(key, value);
 }
 
 #ifndef RABIT_STRICT_CXX98_
@@ -198,7 +199,7 @@ inline void TrackerPrintf(const char *fmt, ...) {
   TrackerPrint(msg);
 }
 
-inline void TrackerSetConfig(const char *key, const int* value, ...) {
+inline void TrackerSetCacheIndex(const char *key, const int* value, ...) {
   const int kPrintBuffer = 1 << 10;
   std::string k(kPrintBuffer, '\0'), v(kPrintBuffer, '\0');
 
@@ -207,10 +208,10 @@ inline void TrackerSetConfig(const char *key, const int* value, ...) {
   vsnprintf(&k[0], kPrintBuffer, key, args1);
   va_end(args1);
   k.resize(strlen(k.c_str()));
-  engine::GetEngine()->TrackerSetConfig(k, *value);
+  engine::GetEngine()->TrackerSetCacheIndex(k, *value);
 }
 
-inline void TrackerGetConfig(const char *key, int* value, ...) {
+inline void TrackerGetCacheIndex(const char *key, int* value, ...) {
   const int kPrintBuffer = 1 << 10;
   std::string k(kPrintBuffer, '\0'), v(kPrintBuffer, '\0');
 
@@ -219,7 +220,7 @@ inline void TrackerGetConfig(const char *key, int* value, ...) {
   vsnprintf(&k[0], kPrintBuffer, key, args1);
   va_end(args1);
   k.resize(strlen(k.c_str()));
-  engine::GetEngine()->TrackerGetConfig(k, *value);
+  engine::GetEngine()->TrackerGetCacheIndex(k, *value);
 }
 #endif  // RABIT_STRICT_CXX98_
 // load latest check point
--- a/include/rabit/internal/rabit-inl.h
+++ b/include/rabit/internal/rabit-inl.h
@@ -178,11 +178,11 @@ inline void TrackerPrint(const std::string &msg) {
   engine::GetEngine()->TrackerPrint(msg);
 }
 
-inline void TrackerSetConfig(const std::string &key, const std::string &value) {
+inline void TrackerSetConfig(const std::string &key, const int &value) {
   engine::GetEngine()->TrackerSetConfig(key, value);
 }
 
-inline void TrackerGetConfig(const std::string &key, std::string* value) {
+inline void TrackerGetConfig(const std::string &key, int &value) {
   engine::GetEngine()->TrackerGetConfig(key, value);
 }
 
@@ -198,36 +198,28 @@ inline void TrackerPrintf(const char *fmt, ...) {
   TrackerPrint(msg);
 }
 
-inline void TrackerSetConfig(const char *key, const char *value, ...) {
+inline void TrackerSetConfig(const char *key, const int* value, ...) {
   const int kPrintBuffer = 1 << 10;
   std::string k(kPrintBuffer, '\0'), v(kPrintBuffer, '\0');
 
-  va_list args1, args2;
+  va_list args1;
   va_start(args1, key);
-  va_start(args2, value);
   vsnprintf(&k[0], kPrintBuffer, key, args1);
-  vsnprintf(&v[0], kPrintBuffer, value, args2);
   va_end(args1);
-  va_end(args2);
   k.resize(strlen(k.c_str()));
-  v.resize(strlen(v.c_str()));
-  engine::GetEngine()->TrackerSetConfig(k, v);
+  engine::GetEngine()->TrackerSetConfig(k, *value);
 }
 
-inline void TrackerGetConfig(const char *key, char* value, ...) {
+inline void TrackerGetConfig(const char *key, int* value, ...) {
   const int kPrintBuffer = 1 << 10;
   std::string k(kPrintBuffer, '\0'), v(kPrintBuffer, '\0');
 
-  va_list args1, args2;
+  va_list args1;
   va_start(args1, key);
-  va_start(args2, value);
   vsnprintf(&k[0], kPrintBuffer, key, args1);
-  vsnprintf(&v[0], kPrintBuffer, value, args2);
   va_end(args1);
-  va_end(args2);
   k.resize(strlen(k.c_str()));
-  v.resize(strlen(v.c_str()));
-  engine::GetEngine()->TrackerGetConfig(k, &v);
+  engine::GetEngine()->TrackerGetConfig(k, *value);
 }
 #endif  // RABIT_STRICT_CXX98_
 // load latest check point
--- a/include/rabit/internal/rabit-inl.h
+++ b/include/rabit/internal/rabit-inl.h
@@ -156,10 +156,9 @@ inline void Broadcast(std::string *sendrecv_data, int root) {
 template<typename OP, typename DType>
 inline void Allreduce(DType *sendrecvbuf, size_t count,
                       void (*prepare_fun)(void *arg),
-                      void *prepare_arg,
-                      int* cache_seq) {
+                      void *prepare_arg) {
   engine::Allreduce_(sendrecvbuf, sizeof(DType), count, op::Reducer<OP, DType>,
-                     engine::mpi::GetType<DType>(), OP::kType, prepare_fun, prepare_arg, cache_seq);
+                     engine::mpi::GetType<DType>(), OP::kType, prepare_fun, prepare_arg);
 }
 
 // C++11 support for lambda prepare function
@@ -168,9 +167,9 @@ inline void InvokeLambda_(void *fun) {
   (*static_cast<std::function<void()>*>(fun))();
 }
 template<typename OP, typename DType>
-inline void Allreduce(DType *sendrecvbuf, size_t count, std::function<void()> prepare_fun, int* cache_seq) {
+inline void Allreduce(DType *sendrecvbuf, size_t count, std::function<void()> prepare_fun) {
   engine::Allreduce_(sendrecvbuf, sizeof(DType), count, op::Reducer<OP, DType>,
-                     engine::mpi::GetType<DType>(), OP::kType, InvokeLambda_, &prepare_fun, cache_seq);
+                     engine::mpi::GetType<DType>(), OP::kType, InvokeLambda_, &prepare_fun);
 }
 #endif  // C++11
 
@@ -179,12 +178,12 @@ inline void TrackerPrint(const std::string &msg) {
   engine::GetEngine()->TrackerPrint(msg);
 }
 
-inline void TrackerSetCacheIndex(const std::string &key, const int &value) {
-  engine::GetEngine()->TrackerSetCacheIndex(key, value);
+inline void SetCache(const std::string &key, const void* buf) {
+  engine::GetEngine()->SetCache(key, buf);
 }
 
-inline void TrackerGetCacheIndex(const std::string &key, int &value) {
-  engine::GetEngine()->TrackerGetCacheIndex(key, value);
+inline void GetCache(const std::string &key, void* buf) {
+  engine::GetEngine()->GetCache(key, buf);
 }
 
 #ifndef RABIT_STRICT_CXX98_
@@ -199,7 +198,7 @@ inline void TrackerPrintf(const char *fmt, ...) {
   TrackerPrint(msg);
 }
 
-inline void TrackerSetCacheIndex(const char *key, const int* value, ...) {
+inline void SetCache(const char *key, const void* buf, ...) {
   const int kPrintBuffer = 1 << 10;
   std::string k(kPrintBuffer, '\0'), v(kPrintBuffer, '\0');
 
@@ -208,10 +207,10 @@ inline void TrackerSetCacheIndex(const char *key, const int* value, ...) {
   vsnprintf(&k[0], kPrintBuffer, key, args1);
   va_end(args1);
   k.resize(strlen(k.c_str()));
-  engine::GetEngine()->TrackerSetCacheIndex(k, *value);
+  engine::GetEngine()->SetCache(k, buf);
 }
 
-inline void TrackerGetCacheIndex(const char *key, int* value, ...) {
+inline void GetCache(const char *key, void* buf, ...) {
   const int kPrintBuffer = 1 << 10;
   std::string k(kPrintBuffer, '\0'), v(kPrintBuffer, '\0');
 
@@ -220,7 +219,7 @@ inline void TrackerGetCacheIndex(const char *key, int* value, ...) {
   vsnprintf(&k[0], kPrintBuffer, key, args1);
   va_end(args1);
   k.resize(strlen(k.c_str()));
-  engine::GetEngine()->TrackerGetCacheIndex(k, *value);
+  engine::GetEngine()->GetCache(k, buf);
 }
 #endif  // RABIT_STRICT_CXX98_
 // load latest check point
--- a/include/rabit/internal/rabit-inl.h
+++ b/include/rabit/internal/rabit-inl.h
@@ -197,6 +197,38 @@ inline void TrackerPrintf(const char *fmt, ...) {
   msg.resize(strlen(msg.c_str()));
   TrackerPrint(msg);
 }
+
+inline void TrackerSetConfig(const char *key, const char *value, ...) {
+  const int kPrintBuffer = 1 << 10;
+  std::string k(kPrintBuffer, '\0'), v(kPrintBuffer, '\0');
+
+  va_list args1, args2;
+  va_start(args1, key);
+  va_start(args2, value);
+  vsnprintf(&k[0], kPrintBuffer, key, args1);
+  vsnprintf(&v[0], kPrintBuffer, value, args2);
+  va_end(args1);
+  va_end(args2);
+  k.resize(strlen(k.c_str()));
+  v.resize(strlen(v.c_str()));
+  engine::GetEngine()->TrackerSetConfig(k, v);
+}
+
+inline void TrackerGetConfig(const char *key, char* value, ...) {
+  const int kPrintBuffer = 1 << 10;
+  std::string k(kPrintBuffer, '\0'), v(kPrintBuffer, '\0');
+
+  va_list args1, args2;
+  va_start(args1, key);
+  va_start(args2, value);
+  vsnprintf(&k[0], kPrintBuffer, key, args1);
+  vsnprintf(&v[0], kPrintBuffer, value, args2);
+  va_end(args1);
+  va_end(args2);
+  k.resize(strlen(k.c_str()));
+  v.resize(strlen(v.c_str()));
+  engine::GetEngine()->TrackerGetConfig(k, &v);
+}
 #endif  // RABIT_STRICT_CXX98_
 // load latest check point
 inline int LoadCheckPoint(Serializable *global_model,
--- a/include/rabit/internal/rabit-inl.h
+++ b/include/rabit/internal/rabit-inl.h
@@ -127,29 +127,28 @@ inline std::string GetProcessorName(void) {
   return engine::GetEngine()->GetHost();
 }
 // broadcast data to all other nodes from root
-inline void Broadcast(void *sendrecv_data, size_t size, int root, const char* caller) {
-  utils::Printf("[%d] Broadcast caller is %s \n", GetRank(), caller);
+inline void Broadcast(void *sendrecv_data, size_t size, int root) {
   engine::GetEngine()->Broadcast(sendrecv_data, size, root);
 }
 template<typename DType>
-inline void Broadcast(std::vector<DType> *sendrecv_data, int root, const char* caller) {
+inline void Broadcast(std::vector<DType> *sendrecv_data, int root) {
   size_t size = sendrecv_data->size();
   Broadcast(&size, sizeof(size), root);
   if (sendrecv_data->size() != size) {
     sendrecv_data->resize(size);
   }
   if (size != 0) {
-    Broadcast(&(*sendrecv_data)[0], size * sizeof(DType), root, caller);
+    Broadcast(&(*sendrecv_data)[0], size * sizeof(DType), root);
   }
 }
-inline void Broadcast(std::string *sendrecv_data, int root, const char* caller) {
+inline void Broadcast(std::string *sendrecv_data, int root) {
   size_t size = sendrecv_data->length();
-  Broadcast(&size, sizeof(size), root, caller);
+  Broadcast(&size, sizeof(size), root);
   if (sendrecv_data->length() != size) {
     sendrecv_data->resize(size);
   }
   if (size != 0) {
-    Broadcast(&(*sendrecv_data)[0], size * sizeof(char), root, caller);
+    Broadcast(&(*sendrecv_data)[0], size * sizeof(char), root);
   }
 }
 
@@ -157,9 +156,7 @@ inline void Broadcast(std::string *sendrecv_data, int root, const char* caller) 
 template<typename OP, typename DType>
 inline void Allreduce(DType *sendrecvbuf, size_t count,
                       void (*prepare_fun)(void *arg),
-                      void *prepare_arg, 
-                      const char* caller) {
-  utils::Printf("[%d] Allreduce caller is %s \n", GetRank(), caller);
+                      void *prepare_arg) {
   engine::Allreduce_(sendrecvbuf, sizeof(DType), count, op::Reducer<OP, DType>,
                      engine::mpi::GetType<DType>(), OP::kType, prepare_fun, prepare_arg);
 }
@@ -170,8 +167,7 @@ inline void InvokeLambda_(void *fun) {
   (*static_cast<std::function<void()>*>(fun))();
 }
 template<typename OP, typename DType>
-inline void Allreduce(DType *sendrecvbuf, size_t count, std::function<void()> prepare_fun, const char* caller) {
-  utils::Printf("[%d] Allreduce caller is %s \n", GetRank(), caller);
+inline void Allreduce(DType *sendrecvbuf, size_t count, std::function<void()> prepare_fun) {
   engine::Allreduce_(sendrecvbuf, sizeof(DType), count, op::Reducer<OP, DType>,
                      engine::mpi::GetType<DType>(), OP::kType, InvokeLambda_, &prepare_fun);
 }
@@ -292,7 +288,7 @@ inline Reducer<DType, freduce>::Reducer(void) {
 template<typename DType, void (*freduce)(DType &dst, const DType &src)> // NOLINT(*)
 inline void Reducer<DType, freduce>::Allreduce(DType *sendrecvbuf, size_t count,
                                                void (*prepare_fun)(void *arg),
-                                               void *prepare_arg, const char* caller) {
+                                               void *prepare_arg) {
   handle_.Allreduce(sendrecvbuf, sizeof(DType), count, prepare_fun, prepare_arg);
 }
 // function to perform reduction for SerializeReducer
@@ -341,9 +337,7 @@ template<typename DType>
 inline void SerializeReducer<DType>::Allreduce(DType *sendrecvobj,
                                                size_t max_nbyte, size_t count,
                                                void (*prepare_fun)(void *arg),
-                                               void *prepare_arg,
-                                               const char* caller) {
-  utils::Printf("[%d] Allreduce caller is %s \n", GetRank(), caller);
+                                               void *prepare_arg) {
   buffer_.resize(max_nbyte * count);
   // setup closure
   SerializeReduceClosure<DType> c;
@@ -361,17 +355,13 @@ inline void SerializeReducer<DType>::Allreduce(DType *sendrecvobj,
 #if DMLC_USE_CXX11
 template<typename DType, void (*freduce)(DType &dst, const DType &src)>  // NOLINT(*)g
 inline void Reducer<DType, freduce>::Allreduce(DType *sendrecvbuf, size_t count,
-                                               std::function<void()> prepare_fun,
-                                               const char* caller) {
-  utils::Printf("[%d] Allreduce caller is %s \n", GetRank(), caller);
+                                               std::function<void()> prepare_fun) {
   this->Allreduce(sendrecvbuf, count, InvokeLambda_, &prepare_fun);
 }
 template<typename DType>
 inline void SerializeReducer<DType>::Allreduce(DType *sendrecvobj,
                                                size_t max_nbytes, size_t count,
-                                               std::function<void()> prepare_fun,
-                                               const char* caller) {
-  utils::Printf("[%d] Allreduce caller is %s \n", GetRank(), caller);
+                                               std::function<void()> prepare_fun) {
   this->Allreduce(sendrecvobj, max_nbytes, count, InvokeLambda_, &prepare_fun);
 }
 #endif  // DMLC_USE_CXX11
--- a/include/rabit/internal/rabit-inl.h
+++ b/include/rabit/internal/rabit-inl.h
@@ -142,7 +142,8 @@ inline void Broadcast(std::vector<DType> *sendrecv_data, int root, const char* c
     Broadcast(&(*sendrecv_data)[0], size * sizeof(DType), root, caller);
   }
 }
-inline void Broadcast(std::string *sendrecv_data, int root, const char* caller) {
+inline void Broadcast(std::string *sendrecv_data, int root, 
+  const char* caller) {
   size_t size = sendrecv_data->length();
   Broadcast(&size, sizeof(size), root, caller);
   if (sendrecv_data->length() != size) {
@@ -182,12 +183,12 @@ inline void TrackerPrint(const std::string &msg) {
   engine::GetEngine()->TrackerPrint(msg);
 }
 
-inline void TrackerSetConfig(const std::string &key, const std::string &value) {
-  engine::GetEngine()->TrackerSetConfig(key, value);
+inline void TrackerSetConfig(const std::string &key, const int bsize, const void *value) {
+  engine::GetEngine()->TrackerSetConfig(key, bsize, value);
 }
 
-inline void TrackerGetConfig(const std::string &key, std::string* value) {
-  engine::GetEngine()->TrackerGetConfig(key, value);
+inline void TrackerGetConfig(const std::string &key, const int bsize, void *value) {
+  engine::GetEngine()->TrackerGetConfig(key, bsize, value);
 }
 
 #ifndef RABIT_STRICT_CXX98_
@@ -202,7 +203,7 @@ inline void TrackerPrintf(const char *fmt, ...) {
   TrackerPrint(msg);
 }
 
-inline void TrackerSetConfig(const char *key, const char *value, ...) {
+inline void TrackerSetConfig(const char *key, const int bsize, const void *value, ...) {
   const int kPrintBuffer = 1 << 10;
   std::string k(kPrintBuffer, '\0'), v(kPrintBuffer, '\0');
 
@@ -210,28 +211,22 @@ inline void TrackerSetConfig(const char *key, const char *value, ...) {
   va_start(args1, key);
   va_start(args2, value);
   vsnprintf(&k[0], kPrintBuffer, key, args1);
-  vsnprintf(&v[0], kPrintBuffer, value, args2);
   va_end(args1);
   va_end(args2);
   k.resize(strlen(k.c_str()));
-  v.resize(strlen(v.c_str()));
-  engine::GetEngine()->TrackerSetConfig(k, v);
+  engine::GetEngine()->TrackerSetConfig(k, bsize, value);
 }
 
-inline void TrackerGetConfig(const char *key, char* value, ...) {
+inline void TrackerGetConfig(const char *key, const int bsize, void* value, ...) {
   const int kPrintBuffer = 1 << 10;
   std::string k(kPrintBuffer, '\0'), v(kPrintBuffer, '\0');
 
-  va_list args1, args2;
+  va_list args1;
   va_start(args1, key);
-  va_start(args2, value);
   vsnprintf(&k[0], kPrintBuffer, key, args1);
-  vsnprintf(&v[0], kPrintBuffer, value, args2);
   va_end(args1);
-  va_end(args2);
   k.resize(strlen(k.c_str()));
-  v.resize(strlen(v.c_str()));
-  engine::GetEngine()->TrackerGetConfig(k, &v);
+  engine::GetEngine()->TrackerGetConfig(k, bsize, value);
 }
 #endif  // RABIT_STRICT_CXX98_
 // load latest check point
--- a/include/rabit/internal/rabit-inl.h
+++ b/include/rabit/internal/rabit-inl.h
@@ -127,28 +127,28 @@ inline std::string GetProcessorName(void) {
   return engine::GetEngine()->GetHost();
 }
 // broadcast data to all other nodes from root
-inline void Broadcast(void *sendrecv_data, size_t size, int root, const char* caller_) {
-  engine::GetEngine()->Broadcast(sendrecv_data, size, root, caller_);
+inline void Broadcast(void *sendrecv_data, size_t size, int root) {
+  engine::GetEngine()->Broadcast(sendrecv_data, size, root);
 }
 template<typename DType>
-inline void Broadcast(std::vector<DType> *sendrecv_data, int root, const char* caller_) {
+inline void Broadcast(std::vector<DType> *sendrecv_data, int root) {
   size_t size = sendrecv_data->size();
   Broadcast(&size, sizeof(size), root);
   if (sendrecv_data->size() != size) {
     sendrecv_data->resize(size);
   }
   if (size != 0) {
-    Broadcast(&(*sendrecv_data)[0], size * sizeof(DType), root, caller_);
+    Broadcast(&(*sendrecv_data)[0], size * sizeof(DType), root);
   }
 }
-inline void Broadcast(std::string *sendrecv_data, int root, const char* caller_) {
+inline void Broadcast(std::string *sendrecv_data, int root) {
   size_t size = sendrecv_data->length();
   Broadcast(&size, sizeof(size), root);
   if (sendrecv_data->length() != size) {
     sendrecv_data->resize(size);
   }
   if (size != 0) {
-    Broadcast(&(*sendrecv_data)[0], size * sizeof(char), root, caller_);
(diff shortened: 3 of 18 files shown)


src/allreduce_robust.h
```diff
@@ -519,6 +579,12 @@ o   *  the input state must exactly one saved state(local state of current node)
   int result_buffer_round;
   // result buffer of all reduce
   ResultBuffer resbuf;
+  // current cached allreduce/braodcast sequence number
+  int cur_cache_seq;
+  // result buffer of cached all reduce
+  ResultBuffer cachebuf;
+  // key of each cache entry
+  ResultBuffer lookupbuf;
```
CodingCat: so it essentially doubles even triples the memory footprint with in an iteration ?

in XGB, we save stats for each node in a tree, and now, we have to save it in rabit layer again?

test/model_recover.cc
```diff
@@ -36,6 +36,7 @@ inline void TestMax(Model *model, int ntrial, int iter) {
     ndata[i] = (i * (rank+1)) % z  + model->data[i];
   }
   rabit::Allreduce<op::Max>(&ndata[0], ndata.size());
+  std::vector<float> ntest(model->data.size());
```
CodingCat: what's this for?

src/allreduce_robust.h
CodingCat: can you give an example PR on XGB side about how to leverage this feature?

test/model_recover.cc
chenqin: removed

src/allreduce_robust.h
chenqin: size of cache is bounded by number of keys user specificly defined to cache (those are not always hits in each iteration like bootstrap or DMatrix loading) 

The way rabit works without cache is it assume each iteration runs exactly same set of allreduce/broadcast operations. If one iteration complete with checkpoint succeed, it reset seq to 0. 

When a node failed, it loads latest checkpoint and reset seq no on that node to 0 which will try to connect to other nodes and align on seq no. As you might imagine, if one side is doing bootstrap and try allreduce with number of colums while others are trying to allreduce with histograms, it wouldn't work.

https://github.com/dmlc/xgboost/pull/4636/files#diff-bc273f37600ddbaef4874e160c1e7ca3R227

chenqin:
covers various of not_always_occurred_allreduce/broadcast_operations as cache usage 

https://github.com/dmlc/rabit/pull/98/files#diff-6eb0ed04b6033c8a6ab6a39d33458a9bR98

> no test for this significant feature?

src/allreduce_robust.h
chenqin: In term of memory footprint overhead, approx method will cost 2 unit64_t bytes, fast histogram will be one hist_init size + 2 unit64_t + a float

CodingCat: > In term of memory footprint overhead, approx method will cost 2 unit64_t bytes, fast histogram will be one hist_init size + 2 unit64_t + a float

this is not true according to https://github.com/dmlc/xgboost/pull/4636/files#diff-daec1d0ffe267847ee82f8b885a06dbbR179. 

every feature will have a sketch and similarly you also need to cache results of histogram corresponding to sketch (which I didn't find in your example PR)

CodingCat:
I am also a bit confused about 

1. if we choose to cache everything, is it even simple to talk with an external KV store instead of having a very simplified distributed kv store here?

2. why we have to cache every step? every worker has built the local checkpoint, it should be very easy to re-do an iteration for all workers

chenqin:
> I am also a bit confused about
> 
> 1. if we choose to cache everything, is it even simpler to talk with an external KV store instead of having a very simplified distributed kv store here? 

If you are talking about rocksdb https://github.com/facebook/rocksdb, yes, I think we should move those ResultBufer to disk and reduce memory footprint. (in a separate pr), we also want to keep unique name of results so that we can merge existing result buffer into cachebuffer and figure out eviction strategy(in a separate pr) If you are talking about something like backed directly by s3, the updates might be an issue. (random lookup and updates within each allreduce/broadcast call) 

> 2. why we have to cache every step? every worker has built the local checkpoint, it should be very easy to re-do an iteration for all workers

we already cache everything per last iteration before this feature. Let's use distributed hist I run as example we cache 14 allreduce/broadcast operation results before rabit clear it. 
During run time we constantly do random lookup and update those values so write to checkpoint file and random access can be slow. (not mention external file system)

CodingCat:
> > I am also a bit confused about
> > 
> > 1. if we choose to cache everything, is it even simpler to talk with an external KV store instead of having a very simplified distributed kv store here?
> 
> If you are talking about rocksdb https://github.com/facebook/rocksdb, yes, I think we should move those ResultBufer to disk and reduce memory footprint. (in a separate pr), we also want to keep unique name of results so that we can merge existing result buffer into cachebuffer and figure out eviction strategy(in a separate pr) If you are talking about something like backed directly by s3, the updates might be an issue. (random lookup and updates within each allreduce/broadcast call)
> 

how about a distributed kv store like HBase which is used to back many web services? (the latency of web services should be much more strict than an offline machine learning model training)

> > 1. why we have to cache every step? every worker has built the local checkpoint, it should be very easy to re-do an iteration for all workers
> 
> we already cache everything per last iteration before this feature. Let's use distributed hist I run as example we cache 14 allreduce/broadcast operation results before rabit clear it.
> During run time we constantly do random lookup and update those values so write to checkpoint file and random access can be slow. (not mention external file system)

no, I am asking that why we have to do additional "cache" instead of using the checkpoint made by each iteration. To me, the following work flow should be very straight forward

worker 1: iter1 -> checkpoint -> iter2 -> re-do from the checkpoint after iter1 (coordinated by tracker)
worker 2: iter1 -> checkpoint -> fail -> re-do from the checkpoint after iter1 (coordinated by tracker)

I think rabit fault-tolerance originally more or less works in such a way

chenqin:
> , I am asking that why we have to do additional "cache" instead of using the checkpoint made by each iteration. To me, the following work flow should be very straight forward

This is how thing works, I mocked rank 0 failure where you can see restarted seqcounter for rank 0 is smaller than rest of nodes.  what you can see if rank 0 recover and load checkpoint then try to run allreduce (seq = 0) where all others do braodcast (seq = 1)

What happens after was other nodes backfill allreduce results back to node 0 so it moves forward to next collective call broadcast (seq = 1)  and catched up with rest of ranks.


`

[0]@@@Hit Mock Error:Broadcast
/home/cq/xgboost/xgboost machine.conf mock=0,10,1,0 DMLC_NUM_ATTEMPT=1
[20:42:25] start instance-xgboost:0
[20:42:25] Load part of data 0 of 4 parts
[20:42:25] 1628x127 matrix with 35816 entries loaded from ../../demo/data/agaricus.txt.train
[0] RecoverExec caller GetCache seq 0
[3] RecoverExec caller Broadcast seq 1
[1] RecoverExec caller Broadcast seq 1
[2] RecoverExec caller Broadcast seq 1
[20:42:25] Load part of data 0 of 4 parts
[20:42:25] 403x127 matrix with 8866 entries loaded from ../../demo/data/agaricus.txt.test
[0] RecoverExec caller GetCache seq 0
[3] RecoverExec caller Broadcast seq 1
[1] RecoverExec caller Broadcast seq 1
[2] RecoverExec caller Broadcast seq 1
[0] RecoverExec caller LoadCheckPoint seq 0
[1] RecoverExec caller Broadcast seq 1
[3] RecoverExec caller Broadcast seq 1
[2] RecoverExec caller Broadcast seq 1
[0] RecoverExec caller LoadCheckPoint seq 0
[0] load checkpoint global 4572 version 10
[1] RecoverExec caller Broadcast seq 1
[3] RecoverExec caller Broadcast seq 1
[2] RecoverExec caller Broadcast seq 1
[0] RecoverExec caller LoadCheckPoint seq 0
[1] RecoverExec caller Broadcast seq 1
[3] RecoverExec caller Broadcast seq 1
[2] RecoverExec caller Broadcast seq 1
[20:42:25] Tree method is automatically selected to be 'approx' for distributed training.
[0] RecoverExec caller Allreduce seq 0
[3] RecoverExec caller Broadcast seq 1
[1] RecoverExec caller Broadcast seq 1
[2] RecoverExec caller Broadcast seq 1
[2] pre trygetresult seqno 0 requester 0
[2] role is 0 in trygetresult seqno 0 seq_counter 1
[3] post tryget result seq_counter 1 requester 0
[1] post tryget result seq_counter 1 requester 0
[0] post tryget result seq_counter 0 requester 1
[2] post tryget result seq_counter 1 requester 0
[0] RecoverExec caller Broadcast seq 1
[3] RecoverExec caller Broadcast seq 1
[1] RecoverExec caller Broadcast seq 1
[2] RecoverExec caller Broadcast seq 1

`

chenqin:
> how about a distributed kv store like HBase which is used to back many web services? (the latency requirement of web services should be much more strict than an offline machine learning model training)

I am hesitate to add a service dependency and less to a library at this point. HBase can be expensive to operate and not everyone will install. I agree this could be one option if people really feel that's how they want to operate.

src/allreduce_robust.h
chenqin: This is runtime log of fast histogram, you can see there are two d_colmn cache call from loading dmatrix and one from seed in colmnsampler, so histogram init only called once.  As long as we would be able to align seq counter in all nodes, we don't need to cache anything further.

`
[0]@@@Hit Mock Error:AllReduce
/home/cq/xgboost/xgboost machinehist.conf mock=0,10,1,0 DMLC_NUM_ATTEMPT=1
[21:13:45] start instance-xgboost:0
[21:13:45] Load part of data 0 of 10 parts
[21:13:45] 650x127 matrix with 14300 entries loaded from ../../demo/data/agaricus.txt.train
[0] RecoverExec caller GetCache seq 0
[21:13:45] Load part of data 0 of 10 parts
[21:13:45] 161x127 matrix with 3542 entries loaded from ../../demo/data/agaricus.txt.test
[0] RecoverExec caller GetCache seq 0
...
[0] RecoverExec caller LoadCheckPoint seq 0
...
[0] RecoverExec caller LoadCheckPoint seq 0
[0] load checkpoint global 4572 version 10
[0] RecoverExec caller LoadCheckPoint seq 0
...
[21:13:45] Tree method is specified to be 'hist' for distributed training.
[0] RecoverExec caller GetCache seq 0
[0] RecoverExec caller GetCache seq 0
...
[0] RecoverExec caller Allreduce seq 0
[1] RecoverExec caller Allreduce seq 1
...
[0] RecoverExec caller Allreduce seq 1
[9] RecoverExec caller Allreduce seq 1
[1] RecoverExec caller Allreduce seq 1
[8] RecoverExec caller Allreduce seq 1
[7] RecoverExec caller Allreduce seq 1
[2] RecoverExec caller Allreduce seq 1
[6] RecoverExec caller Allreduce seq 1
[3] RecoverExec caller Allreduce seq 1
[4] RecoverExec caller Allreduce seq 1
[5] RecoverExec caller Allreduce seq 1
...
`

CodingCat:
> This is how thing works, I mocked rank 0 failure where you can see restarted seqcounter for rank 0 is smaller than rest of nodes. what you can see if rank 0 recover and load checkpoint then try to run allreduce (seq = 0) where all others do braodcast (seq = 1)

I think a more correct way to address is 

1. understand why others starts with seq = 1 even they should restart from the previous checkpoint. 

2. ensure that the seq is reset when the workers are requested from the last checkpoint 

instead of workaround it

src/allreduce_robust.h
CodingCat: I agree it should be called for only once for caching sketch, but 

1. it still nearly doubles the memory footprint and `approx method will cost 2 unit64_t bytes` should not be true, as sketch is nearly the same in these two algorithms, 

2. I am also not convinced about why you didn't cache histogram itself in fast histogram and sketch in approx

chenqin:
> > This is how thing works, I mocked rank 0 failure where you can see restarted seqcounter for rank 0 is smaller than rest of nodes. what you can see if rank 0 recover and load checkpoint then try to run allreduce (seq = 0) where all others do braodcast (seq = 1)
> 
> I think a more correct way to address is
> 
> 1. understand why others start with seq = 1 even they should restart from the previous checkpoint.
> 2. ensure that the seq is reset when the workers are requested from the last checkpoint
> 
> instead of workaround it

rabit paper attached. please check section 3.1 there are 5 steps 
https://pdfs.semanticscholar.org/cc42/7b070f214ad11f4b8e7e4e0f0a5bfa9d55bf.pdf

CodingCat:
> > > This is how thing works, I mocked rank 0 failure where you can see restarted seqcounter for rank 0 is smaller than rest of nodes. what you can see if rank 0 recover and load checkpoint then try to run allreduce (seq = 0) where all others do braodcast (seq = 1)
> > 
> > 
> > I think a more correct way to address is
> > 
> > 1. understand why others start with seq = 1 even they should restart from the previous checkpoint.
> > 2. ensure that the seq is reset when the workers are requested from the last checkpoint
> > 
> > instead of workaround it
> 
> rabit paper attached. please check section 3.1 there are 5 steps
> https://pdfs.semanticscholar.org/cc42/7b070f214ad11f4b8e7e4e0f0a5bfa9d55bf.pdf

I DO NOT understand the logic here

1. the paper describes rabit without anything like cache results of every operation

2. you ask me to read the paper that assumes the paper is correct

3. you posted a PR with caching every operation still asking me to read the paper

so, which one is correct?

chenqin:
> > > > This is how thing works, I mocked rank 0 failure where you can see restarted seqcounter for rank 0 is smaller than rest of nodes. what you can see if rank 0 recover and load checkpoint then try to run allreduce (seq = 0) where all others do braodcast (seq = 1)
> > > 
> > > 
> > > I think a more correct way to address is
> > > 
> > > 1. understand why others start with seq = 1 even they should restart from the previous checkpoint.
> > > 2. ensure that the seq is reset when the workers are requested from the last checkpoint
> > > 
> > > instead of workaround it
> > 
> > 
> > rabit paper attached. please check section 3.1 there are 5 steps
> > https://pdfs.semanticscholar.org/cc42/7b070f214ad11f4b8e7e4e0f0a5bfa9d55bf.pdf
> 
> I DO NOT understand the logic here
> 
> 1. the paper describes rabit without anything like cache results of every operation

if we don't cache previous allreduce results history, how can we resume? the approach you suggested assumes checkpointing every allreduce/broadcast iteration.

> 2. you ask me to read the paper that assumes the paper is correct

cache everything in allreduce in rabit design, we already cache everything within one iteration to help failed node catch up.

> 3. you posted a PR with caching every operation still asking me to read the paper

can you point out how everything is cached? as I showed log result showing only one hist init and three int were cached.

> 
> so, which one is correct?

CodingCat:
> > > > > This is how thing works, I mocked rank 0 failure where you can see restarted seqcounter for rank 0 is smaller than rest of nodes. what you can see if rank 0 recover and load checkpoint then try to run allreduce (seq = 0) where all others do braodcast (seq = 1)
> > > > 
> > > > 
> > > > I think a more correct way to address is
> > > > 
> > > > 1. understand why others start with seq = 1 even they should restart from the previous checkpoint.
> > > > 2. ensure that the seq is reset when the workers are requested from the last checkpoint
> > > > 
> > > > instead of workaround it
> > > 
> > > 
> > > rabit paper attached. please check section 3.1 there are 5 steps
> > > https://pdfs.semanticscholar.org/cc42/7b070f214ad11f4b8e7e4e0f0a5bfa9d55bf.pdf
> > 
> > 
> > I DO NOT understand the logic here
> > 
> > 1. the paper describes rabit without anything like cache results of every operation
> 
> if we don't cache previous allreduce results history, how can we resume? the approach you suggested assumes checkpointing every allreduce/broadcast iteration.

it is checkpointing for each iteration https://github.com/dmlc/xgboost/blob/b374e0a7abfdb5571e8737dad7450f25495530d2/jvm-packages/xgboost4j/src/main/java/ml/dmlc/xgboost4j/java/XGBoost.java#L192

I am suggesting recover from the beginning of the iteration not for a particular operation. 

Forget about this PR and what I suggested, the recovery model of rabit is supposed to be in the granularity of iteration instead of an operation. Because 

1. as you said, we do not checkpoint for each operation 

2. when a node recovers, the only history it can see is a checkpointed booster from the last iteration

3. because of 1 and 2, a node has to start from the beginning of an iteration instead of the beginning of a broadcast/allreduce operation

> 
> > 1. you ask me to read the paper that assumes the paper is correct
> 
> cache everything in allreduce in rabit design, we already cache everything within one iteration to help failed node catch up.

the thing I am still unsure about is how things work in this PR

for example, 

worker 1: allreduce sketch -> cache sketch -> build histogram for node 1 -> build histogram for node 2
worker 2: allreduce sketch -> cache sketch -> build histogram for node 1 -> failed -> recover from cache in this PR -> get sketch -> ????

the place I marked with ???? means "how you get worker 2 to the stage building histogram for node 2"?  I can understand if you are able to load a rabit checkpoint everything looks straightforward to recover from the `cached` operation results, how about the task is scheduled to a machine without a checkpoint?

src/allreduce_robust.cc
```diff
@@ -102,13 +162,7 @@ void AllreduceRobust::Allreduce(void *sendrecvbuf_,
     if (prepare_fun != NULL) prepare_fun(prepare_arg);
     return;
   }
-  bool recovered = RecoverExec(sendrecvbuf_, type_nbytes * count, 0, seq_counter);
-  // now we are free to remove the last result, if any
-  if (resbuf.LastSeqNo() != -1 &&
-      (result_buffer_round == -1 ||
-       resbuf.LastSeqNo() % result_buffer_round != rank % result_buffer_round)) {
-    resbuf.DropLast();
```
CodingCat: this is actually relevant to my confusion, looks like you keep the newly added cached content (like a sketch) in memory forever, otherwise it is impossible to recover after 100s of iterations later...

I am very curious how you do that as the seqno is reused per iteration, is it achieved here? what's the impact of not cleaning the buffer after recovery?

chenqin: It's old code which I haven't changed.  In short, anything relate to resbuf were not changed.  From what I read, resbuf is the storage section where it might store more than seq_counter linked allreduce/broadcast payload. 

The original rabit recovery beyond loadcheckpoint relay on seq_counter alignment. a quick example

worker 0 allreduce -> do sorting -> broadcast (seq_counter = 1, paused because link to worker 1 reset) -> consensus process involving worker 0 and worker 1 -> checkpoint
worker 1 allreduce -> do sorting -> xxxx(dead) -> loadcheckpoint -> allreduce(call RecoverExec with seq_counter = 0) -> consensus process involving worker 0 and worker 1 -> allreduce with copied seq0 allreduce result from worker 0 (detail here https://github.com/dmlc/rabit/pull/98/files#diff-6abf0a70e514e75271b0281b21669e1bR1049) -> do sorting -> broadcast(seq_counter = 1) -> consensus process involving worker 0 and 1 (no seq number difference hence go here https://github.com/dmlc/rabit/pull/98/files#diff-6abf0a70e514e75271b0281b21669e1bR1052) -> fallback to tryallreduce with exactly same boradcast calls in both worker 0 and worker 1) -> from this point, it's fully synced.


https://github.com/dmlc/rabit/pull/98/files#diff-6abf0a70e514e75271b0281b21669e1bR1025

CodingCat: this part of code was removed from the PR, I was asking why removed?

chenqin: seq_diff flag is used to mark if every node has same seq_counter, if so, it means all nodes are synced and no need to recover from resbuf in other nodes.

local and global minimal seq_counter is key mono-incremental counter used to align failed worker recovery. if local seq_counter is global one, it means either local is falling behind and get data from other nodes.

CodingCat: Are you answering my question about “why removed?”

CodingCat:
A more general feedback on this PR is, we complicate an allreduce implementation too much,

As a user of allreduce framework, I should not worry about what is to be from cache so that I should set a cache instead of a checkpoint and what is checkpointed so that I can expect it is recovered as expected

Allreduce should be checkpoint+network operations...

So, if something cannot be recovered from a checkpoint, shall we just add that in to ensure allreduce layer is as clean as possible?

src/allreduce_robust.cc
chenqin: > Are you answering my question about “why removed?”
I guess I removed it in previous commiets, I don't see why broadcast needs to remove last resbuf at first place v.s allreduce not. I can add em back.

chenqin:
> A more general feedback on this PR is, we complicate an allreduce implementation too much,
> 
> As a user of allreduce framework, I should not worry about what is to be from cache so that I should set a cache instead of a checkpoint and what is checkpointed so that I can expect it is recovered not needing anything else
> 
> Allreduce should be checkpoint+network operations...
> 
> So, if something cannot be recovered from a checkpoint, shall we just add that in to ensure allreduce layer is as clean as possible?

I agree with clean divide and conquer principle, also needs to point out rabit has initial design of split checkpoint from result cache. There are good amount of trade off in making those differences.

The goal of this pr is not to rewrite rabit but fix things broken for years in distributed recovery.  As a new feature, I haven't expose it to either c_api nor java side for good reason of stability of failure recovery as well as usability.

chenqin:
> > also needs to point out rabit has initial design of split checkpoint from result cache. There are good amount of trade off in making those differences.
> 
> it's different, result cache in the original rabit is more about an optimization to avoid recompute in the health node and it is completely hidden from the user...if we do not have result cache, we can also achieve the same ft functionality which is coordinate every worker to start from a checkpoint and re-compute

If I understand correctly, It's same as fail all nodes and load checkpoint starts from same code path. in this case checkpoint has to be stored in remote file system since all nodes are failing and restart. If we decided to go for this approach, we can get ride of allreducerobust and link to allreducebase. Checkpoint and loadcheckpoint can be done relatively easily with dmlc::stream pointing to a url path. 
I wasn't familiar how interface between xgb-spark and xgboost was done. How can we make sure xgboost node failure will not cause spark task fail?

> 
> but this PR exposes SetCache/GetCache to XGBoost layer, which means, as I said, when I develop a new tree growing algorithm in xgboost, I need to `worry about what is to be from cache so that I should set a cache instead of a checkpoint and what is checkpointed so that I can expect it is recovered not needing anything else`

That's fair point, another way is to eliminate cache proposed in this pr and keep resbuf and checkpoint payload in rocksdb (disk) and always keep unique list of allreduce/broadcast call payloads identified by signature (hide to user) take fast hist as example, kv store keeps three category of results on disk.

------------------------------
bootstrap section (before first success checkpoint) (essentially used to serve as cache)
'dmatrix::load': 127
'columnsamper::init': 3
'histogram::init' : 'binary'
...
------------------------------
iteration section (after last checkpoint section) (essentially what resbuf query and pushtemp did)
synchisto:seq0: ''
...
buildhistorgram:seq10:''
--------------------------------
last success checkpoint (essentially checkpoint payload)
version x: ''
------------------------------

so the checkpoint will be like any node do async snapshot of all three sections and upload to hdfs while executing next iterations mutating content of table. https://rocksdb.org/blog/2015/11/10/use-checkpoints-for-efficient-snapshots.html


> 
> > fix things broken for years in distributed recovery
> 
> I am not sure about whether it has fixed the issue
> 
> first of all, someone still complain about different errors in xgboost repo based on your changes, and it is not tested with anything in production
> 
> second, I am seeing changes in DMatrix::Load() and nothing in approx algorithm, that means nothing is changed in the current XGB-Spark code path, but obviously XGB-Spark single task recovery is not working...
> 
> third, there is something weird like https://github.com/dmlc/xgboost/pull/4636/files#diff-3d55545d6652c5ed31b10509913e0f48R210 and https://github.com/dmlc/xgboost/pull/4636/files#diff-10aac201acca00858b390ff0d679434fR227

example pr actual works in progress.
> 
> > As a new feature, I haven't expose it to either c_api nor java side for good reason of stability of failure recovery
> 
> I am not talking about XGB users, I am talking about rabit user, so you have exposed new APIs to XGBoost devs to (repeat) `worry about what is to be from cache so that I should set a cache instead of a checkpoint and what is checkpointed so that I can expect it is recovered not needing anything else`
> 
> back to xgboost's usage of allreduce, a quick glance of the PR tells me things that are missing from the original recovery
> 
>     1. column_number: if you loaded a checkpoint of model, it should be recovered from the model itself instead of relying on another allreduce operation
> 
>     2. sketch: for approx, it is recalculated everytime, looks like fine, but for fast-histogram, it reuses the init results, so we need to checkpoint it => obviously, what is to be checkpointed is different for different tree growing algorithms, as a user of rabit, I should not using different APIs for this situation but only change what is passed to SaveRabitCheckpoint()...the logic is that HBase provides a single set of APIs like Put/Get, whether to add random salt in my key is my business logic, but not something I should defer to HBase to demand some API called Put and PutWithHashSalt()

same as rocksdb like, I inline to have rocksdb as lib to deal with consistency and snapshot gen. XGB-Spark might be able to integrate with uploading local binary to remote and vice versa.

CodingCat:
> it's same as fail all nodes and load checkpoint starts from same code path. in this case checkpoint has to be stored in remote file system since all nodes are failing and restart. 

no, my proposal is actually inspired by flink's rollback mechanism http://www.vldb.org/pvldb/vol10/p1718-carbone.pdf (3.2.2)

chenqin:
> > it's same as fail all nodes and load checkpoint starts from same code path. in this case checkpoint has to be stored in remote file system since all nodes are failing and restart.
> 
> no, my proposal is actually inspired by flink's rollback mechanism http://www.vldb.org/pvldb/vol10/p1718-carbone.pdf (3.2.2)

Yes, I read about it before, two key points IMO
1) source can be rewind anytime
2) restart all hosts once single point of failure happens.

CodingCat:
> > > it's same as fail all nodes and load checkpoint starts from same code path. in this case checkpoint has to be stored in remote file system since all nodes are failing and restart.
> > 
> > 
> > no, my proposal is actually inspired by flink's rollback mechanism http://www.vldb.org/pvldb/vol10/p1718-carbone.pdf (3.2.2)
> 
> Yes, I read about it before, two key points IMO
> 
> 1. source can be rewind anytime
> 2. restart all hosts once single point of failure happens.

1. there is no addressable source for machine learning, so as long as the partition data is there , it should be fine

2. we are not to move flink's mechanism to here directly, let's focus on this part `upon initialization, retrieve their allocated shard of state. `, in rabit + xgb context, those steps happened only in the first iteration is treated as `initialization`, the model from the last iteration is `allocated shard of state`...so once there is a failure, tracker can detect that, and then broadcast a command to all workers

this command instructs the workers to load checkpoint (this step actually rollback the model to the last iteration), and then the workers are to run `initialization` and load the result buffer or rerun allreduce/broadcast....

but I found the current code structure makes it very difficult to pull workers back from a running allreduce/broadcast....

chenqin:
> > > > it's same as fail all nodes and load checkpoint starts from same code path. in this case checkpoint has to be stored in remote file system since all nodes are failing and restart.
> > > 
> > > 
> > > no, my proposal is actually inspired by flink's rollback mechanism http://www.vldb.org/pvldb/vol10/p1718-carbone.pdf (3.2.2)
> > 
> > 
> > Yes, I read about it before, two key points IMO
> > 
> > 1. source can be rewind anytime
> > 2. restart all hosts once single point of failure happens.
> 
> 1. there is no addressable source for machine learning, so as long as the partition data is there , it should be fine

that's true

> 2. we are not to move flink's mechanism to here directly, let's focus on this part `upon initialization, retrieve their allocated shard of state. `, in rabit + xgb context, those steps happened only in the first iteration is treated as `initialization`, the model from the last iteration is `allocated shard of state`...so once there is a failure, tracker can detect that, and then broadcast a command to all workers
> 

Just double confirm allreduce and broadcast are synchronous blocking call and results are same across all nodes, unlike partitioned operator keyby some partition key and hold part of state of logic operation. Rabit state is much naive simpler. So if we want to apply what the paper described, we can also do checkpoint or any rank really. 

yeah, we can instrument init in rabit level but logic actually runs on framework like xgboost. Certain amount of in memory objects (user states in flink term) in xgboost needs to recompute from last checkpoint and we just go back to restart strategy.

src/allreduce_robust.h
```diff
@@ -177,64 +199,91 @@ class AllreduceRobust : public AllreduceBase {
...
     // action instead of normal execution
-    static const int kDiffSeq = 8;
+    static const int kDiffSeq = 16;
     // constructor
     ActionSummary(void) {}
     // constructor of action
-    explicit ActionSummary(int flag, int minseqno = kSpecialOp) {
-      seqcode = (minseqno << 4) | flag;
+    explicit ActionSummary(int action_flag, int role_diff_flag = 0,
+      int minseqno = kSpecialOp, int maxseqno = kSpecialOp) {
```
CodingCat: why we need to regulate max/min seq no

CodingCat: I don't think there is something like int overflow, 

if seqno is increment by 1 for each operation, in fast histo,  for depthwise growing, (for iterations after the first one) we only have 1 sync for histogram, for loss-guided, we have at most number of leaf nodes sync; for approx, the number of sync is tree's height, 

why it is overflowing?

src/allreduce_robust.cc
CodingCat: if you don't understand why, could you either figure it out or keep it there?

src/allreduce_robust.h
```diff
@@ -177,64 +199,91 @@ class AllreduceRobust : public AllreduceBase {
     // check point Ack, we use a two phase message in check point,
     // this is the second phase of check pointing
     static const int kCheckAck = 4;
+    // some node want to load cache
+    static const int kLoadCache = 8;
```
CodingCat: can you keep the original flags order and append your new thing ?

src/engine_empty.cc
```diff
@@ -73,6 +73,16 @@ class EmptyEngine : public IEngine {
     // simply print information into the tracker
     utils::Printf("%s", msg.c_str());
   }
+  virtual int SetCache(const std::string &key, const void* buf, const size_t buflen) {
```
CodingCat: if this thing is only for cases that allreduce worker are having different allreduce operations in different stage, I would suggest just remove it 

let's only support the simplest case for now, from https://github.com/dmlc/xgboost/pull/4636/files, there is no calling of setcache for approx algorithm, that means we can rely on the original checkpoint mechanism for the default algorithm

src/allreduce_robust.h
chenqin: >why we need to regulate max seq no 
maxseqno is keeping cache_seq and seperate itself from allreduce seqno, what it does is to get number of cache entries in healthy and decide if backfill recovered host is needed.

>why it is overflowing?
First what overflow is minseqno https://github.com/dmlc/xgboost/issues/4250#issuecomment-510745434. what minseqno does is to keep minimal seqno within all nodes, with one exception. 

kSpecialOp which is very large number  https://github.com/dmlc/rabit/blob/master/src/allreduce_robust.h#L165
when it construct, it bitwise move 4 digits to leave space for four masks, latter move 5 digits as we tried to add one more mask. 

https://github.com/dmlc/rabit/blob/master/src/allreduce_robust.h#L188
This is likely cause overflow as singed int type (1 << 26) << 5 > 2^31-1

src/allreduce_robust.h
chenqin: fixed.

src/allreduce_robust.h
```diff
@@ -177,64 +199,91 @@ class AllreduceRobust : public AllreduceBase {
...
     // there are difference sequence number the nodes proposed
     // this means we want to do recover execution of the lower sequence
     // action instead of normal execution
-    static const int kDiffSeq = 8;
+    static const int kDiffSeq = 16;
```
CodingCat: do we really need to add a bit? maybe make the new operation as 9, and just do XOR to check?

src/engine_empty.cc
CodingCat: ignore this, as I said in xgb repo, it's very complicated actually

src/allreduce_robust.h
chenqin: >if you don't understand why, could you either figure it out or keep it there?

it's there 
https://github.com/dmlc/rabit/pull/98/files/c3d653541eaf0a18023acb20150445161f471cc7#diff-6abf0a70e514e75271b0281b21669e1bR195

src/allreduce_robust.h
chenqin: those bits could exists at same time 2 | 4 | 8 | 16
XOR lose it's property as reducer might run multiple times on different nodes

chenqin: humm, a min

src/allreduce_robust.cc
chenqin: fixed, adding back both broadcast and allreduce

src/allreduce_robust.h
CodingCat: why the relationship is defined as `min` and `max`? you just distinguish operations from different sources

same for SeqType, why it is called OR/AND ?

src/allreduce_robust.h
CodingCat: I searched code base a bit, I didn't see the different bits in the flag can be set at the same time? can you point me to the code if there is any?

src/allreduce_robust.h
```diff
@@ -155,18 +172,23 @@ class AllreduceRobust : public AllreduceBase {
...
   struct ActionSummary {
     // maximumly allowed sequence id
-    static const int kSpecialOp = (1 << 26);
+    static const int kSpecialOp = (1 << 20);
     // special sequence number for local state checkpoint
-    static const int kLocalCheckPoint = (1 << 26) - 2;
+    static const int kLocalCheckPoint = (1 << 20) - 2;
     // special sequnce number for local state checkpoint ack signal
-    static const int kLocalCheckAck = (1 << 26) - 1;
+    static const int kLocalCheckAck = (1 << 20) - 1;
```
CodingCat: why we change it to 20?

CMakeLists.txt
```diff
@@ -1,10 +1,10 @@
-cmake_minimum_required(VERSION 3.0)
+cmake_minimum_required(VERSION 3.3)
 
-project(rabit VERSION 0.2.0)
+project(rabit VERSION 0.2.1)
 
 option(RABIT_BUILD_TESTS "Build rabit tests" OFF)
 option(RABIT_BUILD_MPI "Build MPI" OFF)
-option(RABIT_BUILD_DMLC "Include DMLC_CORE in build" ON)
+option(RABIT_BUILD_DMLC "Include DMLC_CORE in build" OFF)
```
CodingCat: why this PR involves changes on make/cmake files?

src/allreduce_robust.h
chenqin: okay, I am afraid we can't use 9, action summary will go through each bit and try to apply either & or | to get permutation of flag sets (1111, 1101 etc). 
      1
    10
  100
1000

About do XOR, I have same thoughts when I tried to find if all nodes are calling getcache at same time. Here is what it might happen, assume four nodes with load_cache bit <1, 1, 0, 0> the expected results would be 1. if we apply xor in reducer, it will produce 0 same as <0,0,0,0> or <1,1,1,1>

CMakeLists.txt
chenqin: this is a bug fix, set on means we expect dmlc-core subdirectory in rabit. It is the case for travis ci as we do run script and download project. In xgboost project, rabit by default should use dmlc-core subproject in same directory level as rabit.

src/allreduce_robust.h
chenqin: so we change to 20, actually I plan to use u_int32_t and keep original special consts https://github.com/chenqin/rabit/commit/b71f662b31dc5867ebb526ce552a6c8120fe1e52

>This is likely cause overflow as singed int type (1 << 26) << 5 + 2^6-1(flags) > 2^31-1

chenqin:
>I searched code base a bit, I didn't see the different bits in the flag can be set at the same time (except the big indicating difference) ? can you point me to the code if there is any?

1.  checking if all nodes have same seqno and recover the node with same seqno as minimal seqno from act (allreduce acknowledgement)
https://github.com/chenqin/rabit/blob/master/src/allreduce_robust.cc#L1095

2. collective checkpointing, if one or more nodes are calling checkpointing by setting checkpoint flag as well as seqno to kSpecialOp, if others are calling same thing, all nodes can do checkpoint at same time. Otherwise, it might indicate at least one node is falling behind calling checkpoint, more comments in code https://github.com/chenqin/rabit/blob/master/src/allreduce_robust.cc#L1019

https://github.com/chenqin/rabit/blob/master/src/allreduce_robust.cc#L1016 

3. if all nodes are calling load checkpoint no one can offer checkpoint, return false
https://github.com/chenqin/rabit/blob/master/src/allreduce_robust.cc#L1064

according to this comments https://github.com/dmlc/rabit/pull/98#discussion_r305127627

maxseqno diff_seq is not used for now. What we can use is to check if everyone have same max cache seq, there is no need to run cache restore at all.

I guess the question is why we introduce this and how it functions.  
Let me break down to what we have before this pr and how  action summary works as single most important primitives in rabit. We want to have quick way to reason which state each node in by calling allreduce on actionsummary. 

ActionSummary act do **minimal** seqno of all nodes
https://github.com/chenqin/rabit/blob/master/src/allreduce_robust.h#L266
ActionSummary act do **or** relationships to all nodes on each flag bits.
https://github.com/chenqin/rabit/blob/master/src/allreduce_robust.h#L268

Two requirements that are slicely different than seqno mimimal / or opeations in action summary reducer function.

As Cache cross seqno reset lifecyles and we want to only restore cache if **not all** nodes are calling getcache. It translate to bit operation of AND (1,1,1,1) = 1 (1,0,1,1) = 0

As we want to recover node without any cache with all cache entries, all nodes needs to know how many cache entreis with "max" value. Those nodes with max value are offering data to those don't. In case of nodes get different cache entries, the node with largest cache entry win (immutable) and rest will get reset to cache entries same as winner. 

Another question is why we need this max instead of use min, the answer lay on an edge case: if node recover, its cache is empty so it will always be min with cache_seq = 0, it doesn't means the rest of nodes actually have anything to offer(they may also be 0) with "max" we can decide if we need to do cache recover at all. 

In future, if we can only recover selectively based on some partition or hashing scheme. (This is one of feature in mind, still lots of questions how to implement. But high level would be if we can partition and replicated DMatrix, we can recover host without reload data from hdfs per sea)

On top of everything we talked about, we want to optimize this consensus allreduce call to minimal payload in one call verses mulitple adhoc tryallreduce calls to get "max" or "or" in flags. That leads to two seperate integers and flags SeqType. I guess naming is confusing at this point.

chenqin:
Overall, I think lots of confusion caused by what's missing in this PR is a detailed design doc and a bit of options we choose. I plan to work on and get everyone on same page.  Stay tune.

CMakeLists.txt
CodingCat: but the build in xgboost is always succeeded?

src/allreduce_robust.h
CodingCat: I mean the choice of 20 is pretty random, if it is just 1 bit added in flag, we didn't choose to 25?

src/allreduce_robust.cc
```diff
@@ -857,18 +971,27 @@ AllreduceRobust::TryGetResult(void *sendrecvbuf, size_t size, int seqno, bool re
...
+  if (flag != 0 && flag != ActionSummary::kLoadCache) {
     utils::Assert(seqno == ActionSummary::kSpecialOp, "must only set seqno for normal operations");
   }
-  // request
-  ActionSummary req(flag, seqno);
+
+  std::string msg = std::string(caller) + " pass negative seqno "
+    + std::to_string(seqno) + " flag " + std::to_string(flag)
+    + " version " + std::to_string(version_number);
+  utils::Assert(seqno >=0, msg.c_str());
```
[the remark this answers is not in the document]
chenqin: fixed

src/allreduce_robust.cc
```diff
@@ -857,18 +971,27 @@ AllreduceRobust::TryGetResult(void *sendrecvbuf, size_t size, int seqno, bool re
  *           result by recovering procedure, the action is complete, no further action is needed
  *    - false means this is the lastest action that has not yet been executed, need to execute the action
  */
-bool AllreduceRobust::RecoverExec(void *buf, size_t size, int flag, int seqno) {
-  if (flag != 0) {
+bool AllreduceRobust::RecoverExec(void *buf, size_t size, int flag, int seqno,
+                                  int cache_seqno, const char* caller) {
+  // skip load cache state as we isolated with assertions
+  if (flag != 0 && flag != ActionSummary::kLoadCache) {
```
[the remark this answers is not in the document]
chenqin: fixed.(reverted)

CMakeLists.txt
chenqin: reverted

src/allreduce_robust.h
chenqin: fixed, use uint32_t and keep offset as it were

src/allreduce_robust.h
chenqin: It was documented in detail in design doc. Simply put, cache behavior is different from normal seq no based catch up.
https://github.com/dmlc/xgboost/files/3482936/Rabit.bootstrap.operations.support.pdf

tqchen:
NOTE, I hade a discussion with @chenqin in person and thinks the general architecture makes sense. But would still be great to have more reviewers to review the code changes.

chenqin:
@hcho3 @terrytangyuan

chenqin:
@CodingCat @hcho3 ping

src/allreduce_robust.cc
```diff
@@ -79,6 +83,68 @@ void AllreduceRobust::SetParam(const char *name, const char *val) {
...
+    }
+  }
+  utils::Assert(index == -1, "immutable cache key already exists");
+  utils::Assert(type_nbytes*count > 0, "can't set empty cache");
+  void* temp = cachebuf.AllocTemp(type_nbytes, count);
+  cachebuf.PushTemp(cur_cache_seq, type_nbytes, count);
+  std::memcpy(temp, buf, type_nbytes*count);
+
+  std::string k(key);
+  void* name = lookupbuf.AllocTemp(strlen(k.c_str()) + 1, 1);
```
hcho3:
```suggestion
  void* name = lookupbuf.AllocTemp(k.length() + 1, 1);
```

src/allreduce_robust.cc
```diff
@@ -79,6 +83,68 @@ void AllreduceRobust::SetParam(const char *name, const char *val) {
...
+  lookupbuf.PushTemp(cur_cache_seq, strlen(k.c_str()) + 1, 1);
```
hcho3:
```suggestion
  lookupbuf.PushTemp(cur_cache_seq, k.length() + 1, 1);
```

src/allreduce_robust.cc
```diff
@@ -79,6 +83,68 @@ void AllreduceRobust::SetParam(const char *name, const char *val) {
...
+  std::memcpy(name, key.c_str(), strlen(k.c_str()) + 1);
```
hcho3:
```suggestion
  std::memcpy(name, key.c_str(), k.length() + 1);
```

src/allreduce_robust.cc
```diff
@@ -79,6 +83,68 @@ void AllreduceRobust::SetParam(const char *name, const char *val) {
...
+int AllreduceRobust::GetCache(const std::string &key, void* buf,
+  const size_t type_nbytes, const size_t count, const bool byref) {
+  // as requester sync with rest of nodes on latest cache content
+  if (!RecoverExec(NULL, 0, ActionSummary::kLoadCache, seq_counter, cur_cache_seq)) return -1;
+
+  int index = -1;
+  for (int i = 0 ; i < cur_cache_seq; i++) {
+    size_t nsize = 0;
+    void* name = lookupbuf.Query(i, &nsize);
+    if (nsize == strlen(key.c_str()) + 1
```
hcho3:
```suggestion
    if (nsize == key.length() + 1
```

src/allreduce_robust.cc
```diff
@@ -79,6 +83,68 @@ void AllreduceRobust::SetParam(const char *name, const char *val) {
...
+      && strcmp(reinterpret_cast<char*>(name), key.c_str()) == 0) {
```
hcho3:
```suggestion
      && strcmp(static_cast<const char*>(name), key.c_str()) == 0) {
```

src/allreduce_robust.cc
```diff
@@ -90,25 +156,44 @@ void AllreduceRobust::SetParam(const char *name, const char *val) {
...
+                                const char* _caller) {
   // skip action in single node
   if (world_size == 1 || world_size == -1) {
     if (prepare_fun != NULL) prepare_fun(prepare_arg);
     return;
   }
-  bool recovered = RecoverExec(sendrecvbuf_, type_nbytes * count, 0, seq_counter);
-  // now we are free to remove the last result, if any
+
+  // genreate unique allreduce signature
```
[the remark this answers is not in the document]
hcho3: Actually, I think that, without unique keys, we won't be able to perform more than one operations in the bootstrap stage, since calling SetCache() twice with the same key will throw an assertion.

So maybe we should simply throw an error when the user attempts to use bootstrap in non-Linux OS?

src/allreduce_base.h
```diff
@@ -525,6 +540,12 @@ class AllreduceBase : public IEngine {
   utils::TCPSocket sock_listen;
   // backdoor port
   int port = 0;
+  // enable bootstrap cache 0 false 1 true
```
hcho3: Is Rabit cache used only for bootstrap? If so, we should rename `rabit_cache` to `rabit_bootstrap_cache`.

src/allreduce_robust.cc
```diff
@@ -79,6 +83,68 @@ void AllreduceRobust::SetParam(const char *name, const char *val) {
     num_local_replica = atoi(val);
   }
 }
+
+int AllreduceRobust::SetCache(const std::string &key, const void *buf,
```
hcho3: Are `SetCache()` and `GetCache()` only used for bootstrap stage? If so, let's rename them to `SetBootstrapCache()` and `GetBootstrapCache()` respectively.

src/allreduce_base.cc
```diff
@@ -283,6 +309,10 @@ bool AllreduceBase::ReConnectLinks(const char *cmd) {
     Assert(rank == -1 || newrank == rank,
            "must keep rank to same if the node already have one");
     rank = newrank;
+
+    // tracker got overwhelemed and not able to assign correct rank
```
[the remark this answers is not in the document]
chenqin: yes, we observed this in running large cluster. New task get -1 rank and still communicate with rest.

chenqin:
> > Overall, I think lots of confusion caused missing a detailed design doc
> 
> @chenqin Can you attach your design doc to this PR?

updated description with link to  [design](https://github.com/dmlc/xgboost/files/3482936/Rabit.bootstrap.operations.support.pdf)

Makefile
```diff
@@ -71,6 +35,23 @@ ifndef LINT_LANG
 	LINT_LANG="all"
 endif
 
+OPENMP_FLAGS =
+ifeq ($(USE_OPENMP), 1)
+	OPENMP_FLAGS = -fopenmp
+else
+	OPENMP_FLAGS = -DDISABLE_OPENMP
```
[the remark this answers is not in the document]
chenqin: removed openmp from this pr, the rationale is we have large number of reduce loop through and we might get benefit of parallel or vectorize oneway or another
`
xgboost_build/src/tree/updater_histmaker.cc::739::CreateHist#16x27354624
`

include/rabit/internal/rabit-inl.h
```diff
@@ -96,6 +100,8 @@ template<typename OP, typename DType>
 inline void Reducer(const void *src_, void *dst_, int len, const MPI::Datatype &dtype) {
   const DType *src = (const DType*)src_;
   DType *dst = (DType*)dst_;  // NOLINT(*)
+
+  #pragma omp parallel for schedule(guided)
```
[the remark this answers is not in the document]
chenqin: yes, removed openmp from this pr.

include/rabit/internal/rabit-inl.h
```diff
@@ -217,6 +250,7 @@ inline void ReducerSafe_(const void *src_, void *dst_, int len_, const MPI::Data
   const char *psrc = reinterpret_cast<const char*>(src_);
   char *pdst = reinterpret_cast<char*>(dst_);
   DType tdst, tsrc;
+  #pragma omp parallel for schedule(guided)
```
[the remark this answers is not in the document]
chenqin: removed

src/allreduce_base.h
chenqin: yes, it's only for bootstrap usage

src/allreduce_robust.cc
chenqin: yes, fixed.

src/allreduce_base.cc
```diff
@@ -211,6 +228,15 @@ void AllreduceBase::SetParam(const char *name, const char *val) {
       throw std::runtime_error("invalid value of DMLC_WORKER_STOP_PROCESS_ON_ERROR");
     }
   }
+  if (!strcmp(name, "rabit_cache")) {
+    rabit_cache = atoi(val);
```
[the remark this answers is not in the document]
chenqin: we do check if(rabit_bootstrap_cache) so if it's negative,user still opt-in this feature
only when rabit_bootstrap_cache default (0), it will not insert/fetch bootstrapcache

chenqin:
Quick summary from rabit user prospective 
- rabit_cache=0 is disabled by default 
- rabit_cache =1 worker will print alert [EXPERIMENTAL] 
- rabit_debug =1 worker will print detailed synchronization logs 
- rabit_reduce_ring_mincount=1000 shows good benefit of improve training spec of deep trees

src/allreduce_base.cc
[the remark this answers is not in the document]
chenqin: This is testing pipeline output, we manually kill mulitple host and wait for retry worker catchup. Notice [-1] come from rabit_debug=1 -1 is default rank.

2019-08-05 18:55:55 INFO  TransportClientFactory:267 - Successfully created connection to agent3872-dca1/10.23.9.172:7337 after 0 ms (0 ms spent in bootstraps)
[-1] loadcheckpoint reset
[-1] checkpoint size 0 finished version 1,seq 0, take 0.000009 seconds
[-1] checkpoint ack finished version 1, take 0.000002 seconds
[-1] checkpoint size 0 finished version 2,seq 0, take 0.000012 seconds
[-1] checkpoint ack finished version 2, take 0.000002 seconds
[-1] checkpoint size 0 finished version 3,seq 0, take 0.000030 seconds
[-1] checkpoint ack finished version 3, take 0.000005 seconds
[-1] checkpoint size 0 finished version 4,seq 0, take 0.000009 seconds
[-1] checkpoint ack finished version 4, take 0.000003 seconds

chenqin:
@hcho3

hcho3:
@chenqin Can we simply throw error when the user is not using Linux and enables bootstrapping? We can add non-Linux support later.

chenqin:
> @chenqin Can we simply throw error when the user is not using Linux and enables bootstrapping? We can add non-Linux support later.

Per convo offline, it still works even with env not supporting _buildin_func, thanks to the fact that we also use type_size and count as part of signature as well.

image

If there is a collision of signature as we change implementation or adding new tree method we could do 

1. explicit change default value and avoid colision
image

2. if we forgot to do so, it will trigger assertion error 
image

include/rabit/internal/utils.h
```diff
@@ -40,6 +40,40 @@
...
+#ifdef __has_builtin
+
+#if __has_builtin(__builtin_FILE)
+#define _FILE  __builtin_FILE()
+#else
+#define _FILE  "N/A"
+#endif  // __has_builtin(__builtin_FILE)
+
+#if __has_builtin(__builtin_LINE)
+#define _LINE  __builtin_LINE()
```
CodingCat: does that mean if we wrap init boostrap in a helper function and call the function with a different higher level caller, it will fail?

src/allreduce_base.cc
```diff
@@ -13,6 +13,10 @@
 #include <cstring>
 #include "./allreduce_base.h"
 
+#ifdef _OPENMP
+#include <omp.h>
+#endif  // _OPENMP
+
```
CodingCat: since open mp has removed, but a lot of headers like this are still here?

include/rabit/internal/rabit-inl.h
CodingCat: it is still using omp?

include/rabit/internal/rabit-inl.h
```diff
@@ -231,6 +265,7 @@ inline void ReducerAlign_(const void *src_, void *dst_,
                           int len_, const MPI::Datatype &dtype) {
   const DType *psrc = reinterpret_cast<const DType*>(src_);
   DType *pdst = reinterpret_cast<DType*>(dst_);
+  #pragma omp parallel for schedule(guided)
```
CodingCat: it is still using omp?

src/allreduce_base.cc
chenqin: fixed.

include/rabit/internal/rabit-inl.h
chenqin: outdated, already removed.

include/rabit/internal/rabit-inl.h
chenqin: outdated, already removed.

include/rabit/internal/utils.h
chenqin: if you were referring someone wrap boostrap allreduce in helper function and call it more than once. It might **fail** but not unexpected behavior.  e.g Dmatrix::Load of training and testing datasets which has exactly same signature. detail how signature were calculated below.

image

when user see failure to run tests, it can follow following way to resolve (pass down unique key from upper function )
image

chenqin: update we remove this in upcoming pr
https://github.com/dmlc/rabit/pull/102/commits/69789ec4510d019d25127c62973acef10d10c6fc

chenqin:
removed is_bootstrap parameter in upcoming pr
https://github.com/dmlc/rabit/pull/102/commits/69789ec4510d019d25127c62973acef10d10c6fc

PR #102, opened 2019-09-03
remove is_bootstrap parameter

chenqin:
remove is_bootstrap parameter from rabit api 
implicit use signature and loadcheckpoint flag instead


chenqin:
@hcho3 @trivialfis @CodingCat this is minor pr remove is_bootstrap parameter from rabit api to make it more user friendly.  https://github.com/dmlc/xgboost/pull/4808 would benefit by having a smaller code change.

src/allreduce_robust.h
```diff
@@ -643,6 +639,8 @@ o   *  the input state must exactly one saved state(local state of current node)
   std::string local_chkpt[2];
   // version of local checkpoint can be 1 or 0
   int local_chkpt_version;
+  // if checkpoint were loaded, used to split results boostrap cache with seqno cache
```
CodingCat: split -> distinguish?

chenqin: fixed.

chenqin:
> could you complete the description about when we load bootstrap cache now?

goog catch, let me update documentation as well.

chenqin:
@CodingCat done

PR #106, opened 2019-09-17
cover change with gtest unittests, clean up cmake script and code includes

chenqin:
This PR do two things
- clear unused environment variables, accept those parameters as arguments 
- add gtest unittests, cleanup includes and cmake scripts to work with xgboost project

@trivialfis @hcho3


src/allreduce_base.cc
```diff
@@ -241,11 +236,7 @@ utils::TCPSocket AllreduceBase::ConnectTracker(void) const {
         utils::Socket::Error("Connect");
       } else {
         fprintf(stderr, "retry connect to ip(retry time %d): [%s]\n", retry, tracker_uri.c_str());
-#if defined(_MSC_VER) || defined (__MINGW32__)
-        Sleep(retry << 1);
-#else
-        sleep(retry << 1);
-#endif
+        sleep(std::rand()%30 + 1);
```
trivialfis: I prefer avoiding random as much as possible.

src/allreduce_base.cc
```diff
@@ -39,16 +42,8 @@ AllreduceBase::AllreduceBase(void) {
   err_link = NULL;
   dmlc_role = "worker";
   this->SetParam("rabit_reduce_buffer", "256MB");
-  // setup possible enviroment variable of intrest
-  env_vars.push_back("rabit_task_id");
```
trivialfis: Is there any possibility of setting up unit tests in rabit?  Even if you can explain everything to me this time, without tests next time someone (might just be me) will still have no clue on what's the result of changes.

hcho3:
@chenqin Can we use exponential backoff to ease congestion? Unlike random waiting, it will have deterministic behavior.

chenqin:
> @chenqin Can we use exponential backoff to ease congestion? Unlike random waiting, it will have removed random with expoential back off plus rank %10 to avoid ddos tracker

chenqin:
> Please use deterministic waiting and point me to some unittests.

working on googletest to support unittests

src/allreduce_base.cc
chenqin: fixed with deterministic backoff

src/allreduce_base.cc
chenqin: added unit tests covering this part.

trivialfis:
@chenqin You can consider using gtest main: https://github.com/dmlc/xgboost/blob/master/tests/cpp/test_main.cc

chenqin:
> @chenqin You can consider using gtest main: https://github.com/dmlc/xgboost/blob/master/tests/cpp/test_main.cc

done

chenqin:
@hcho3 @trivialfis

chenqin:
> Will give a proper review tomorrow, one thing though, try find_package first then download gtest ...

I tried to use find_package along with exteranl project which doesn't ends well. 
gtest download is shelled with RABIT_BUILD_TESTS

trivialfis:
@chenqin Let me try.

trivialfis:
Also, why is it necessary to download gtest?  Can we assume that the developer can install gtest test somewhere?  CMake accepts a parameter called `GTEST_ROOT`, which can be used to find gtest: https://cmake.org/cmake/help/v3.0/module/FindGTest.html

trivialfis:
@chenqin Will make a PR for cleaning up CMake scripts.

chenqin:
@trivialfis let's get this in to unblock xgb CMake house cleaning

chenqin:
done with cleanup, xgboost now build with rabit

src/allreduce_base.cc
```diff
@@ -241,11 +233,7 @@ utils::TCPSocket AllreduceBase::ConnectTracker(void) const {
         utils::Socket::Error("Connect");
       } else {
         fprintf(stderr, "retry connect to ip(retry time %d): [%s]\n", retry, tracker_uri.c_str());
-#if defined(_MSC_VER) || defined (__MINGW32__)
-        Sleep(retry << 1);
-#else
-        sleep(retry << 1);
-#endif
+        sleep((1 << retry) + rank%10);
```
CodingCat: I don't think it resolves problem, if we have 1000 workers, still 1, 11, 21, 31....101, 111, 121....901, 911..., 991 fire connection at the same time

CodingCat: expotential backoff doesn't work like this at all, 

retry 1 : (2^1)
retry 2: choose among (2^1, 2^2)
retry 3: choose among (2^1, 2^2, 2^3)

it is where randomness comes from

chenqin: >I don't think it resolves problem, if we have 1000 workers, still 1, 11, 21, 31....101, 111, 121....901, 911..., 991 fire connection at the same time

yes, this reduce concurrent reconnect from 1000 to 100 every second, assuming largest cluster is less than 2k and it seems fine for tracker to handle 200 hosts job running and reconnect in production.

> expotential backoff doesn't work like this at all

Yes, the problem is exponential back off by definition is random choice.

CodingCat: we should leverage the randomness in exponential backoff instead of inventing by ourselves to avoid congestion network

chenqin:
> we should avoid mixing things together, especially "cleanup/refactoring" and "new feature adding"
> 
> and the other thing I don't understand is. the description says "clear unused environment variables, accept those parameters as arguments" why it contains so many CMake file changes

removed from this pr, let's do back off in separate pr

src/allreduce_base.cc
chenqin: removed from this pr

chenqin:
> we should avoid mixing things together, especially "cleanup/refactoring" and "new feature adding"
> 
> and the other thing I don't understand is. the description says "clear unused environment variables, accept those parameters as arguments" why it contains so many CMake file changes

fixed, remvoed code change touch sleep, this pr focus on make it buildable with xgboost project and adding googletest backed unit tests

chenqin:
@trivialfis

cmake/googletest-download.cmake
```diff
@@ -0,0 +1,20 @@
+# code copied from https://crascit.com/2015/07/25/cmake-gtest/
```
trivialfis: Doesn't seem necessary anymore.  I'm OK for keeping it here for the convenience of others.

test/cpp/allreduce_base_test.cpp
```diff
@@ -0,0 +1,66 @@
+#define RABIT_CXXTESTDEFS_H
```
trivialfis: Awesome, now we have some rigorous unittesting for Rabit.

CMakeLists.txt
```diff
@@ -1,85 +1,106 @@
...
+project(rabit VERSION 0.3.0 LANGUAGES CXX)
 
 option(RABIT_BUILD_TESTS "Build rabit tests" OFF)
 option(RABIT_BUILD_MPI "Build MPI" OFF)
 option(RABIT_BUILD_DMLC "Include DMLC_CORE in build" OFF)
+option(RABIT_WITH_R_LIB "Fit the strict environment of R" OFF)
+
+option(DMLC_ROOT "Specify root of external dmlc core.")
+# by default point to xgboost/dmlc-core
+set(DMLC_ROOT ${CMAKE_CURRENT_LIST_DIR}/../dmlc-core)
```
trivialfis: You spent so much effort on rabit I assume you don't want to make it XGBoost exclusive right?  Since last time I check rabit is only using some headers from `dmlc-core`, there's no linkage involved.  So technically rabit can use dmlc-core in system path, say `/usr/` or `/home/chenqin/.local/` or any other directories that follow the `bin/include/lib/lib64` pattern without any problem, and you can set above path as `DMLC_ROOT` for you convenience.  With this, you can simply throw an error if DMLC_ROOT is not specified.

In the future (not for this PR), I may add this to rabit as `dmlc-core` has very good support for CMake target.
``` cmake
if (NOT DMLC_ROOT)
  find_package(dmlc)
  target_link_libraries(rabit dmlc::dmlc)  # This line also handles include directory.
endif (NOT DMLC_ROOT)
```

CMakeLists.txt
```diff
@@ -1,85 +1,106 @@
...
+  set_target_properties(rabit rabit_base rabit_empty rabit_mock rabit_mock_static
+    PROPERTIES CXX_STANDARD 11
+    CXX_STANDARD_REQUIRED ON
+    POSITION_INDEPENDENT_CODE ON)
 ENDIF(R_LIB OR MINGW OR WIN32)
 
 if(RABIT_BUILD_MPI)
   find_package(MPI REQUIRED)
+  if (NOT MPI_CXX_FOUND)
+    message(FATAL_ERROR "CXX Interface for MPI is required for building MPI backend.")
```
trivialfis: I wrote this line, but do we actually want to support MPI backend?  I thought it's a proof of concept for the paper.  And now that the MPI CXX interface is deprecated, there will be some more works for bring it up. @chenqin What do you think?

chenqin: We might discuss this in another thread how to support MPI runtime. The rationale is there are lots of similar frameworks highly optimized such as facebook/gloo that just do MPI optimization. Not of which offers failed single worker recover. But I don't have clear picture who is using rabit other than distributed XGBoost.

cmake/googletest-download.cmake
hcho3: @trivialfis I added this because this was the easiest way to install Google Test on Windows platform.

PR #108, opened 2019-09-19
remove unreached code in AllreduceRobust::CheckAndRecover

sperlingxx:


trivialfis:
@chenqin Feels like spring cleaning for rabit.

chenqin:
Yeah, just live a notes here. The way it restore connection is problematic when you have large cluster. We might needs to put more thoughts on how to deal with single worker recovery without swamp tracker...

PR #111, opened 2019-09-29
unittests mock, cleanup

chenqin:
cleanup pr following removal of is_bootstrap parameter in https://github.com/dmlc/rabit/commit/9a7ac85d7eb65b1a0b904e1fa8d5a01b910adda4

boostrap_cache no longer support after first loadcheckpoint called, we can reuse seq_no diff with load_cache to see if a worker in recover mode.

support spark mock failure, not exit process

add unittests on mock interface

@trivialfis  @hcho3 @CodingCat

### DIFF
--- a/src/allreduce_robust.h
+++ b/src/allreduce_robust.h
@@ -49,7 +49,7 @@ class AllreduceRobust : public AllreduceBase {
    * \param buflen total number of bytes
    */
   int GetBootstrapCache(const std::string &key, void *buf, const size_t type_nbytes,
-    const size_t count, const bool byref = false);
+    const size_t count);
   /*!
    * \brief perform in-place allreduce, on sendrecvbuf
    *        this function is NOT thread-safe
@@ -255,9 +255,8 @@ class AllreduceRobust : public AllreduceBase {
       return (code & kCheckAck) != 0;
     }
     // whether the operation set contains different sequence number
-    inline bool diff_seq(SeqType t = SeqType::kSeq) const {
-      int code = t == SeqType::kSeq ? seqcode : maxseqcode;
-      return (code & kDiffSeq) != 0;
+    inline bool diff_seq() const {
+      return (seqcode & kDiffSeq) != 0;
     }
     // returns the operation flag of the result
     inline int flag(SeqType t = SeqType::kSeq) const {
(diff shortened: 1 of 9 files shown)


CodingCat:
waiting for CI

chenqin:
@hcho3 @trivialfis can we merge this

test/model_recover.cc
```diff
@@ -96,7 +96,7 @@ int main(int argc, char *argv[]) {
   std::string name = rabit::GetProcessorName();
 
   int max_rank = rank;
-  rabit::Allreduce<op::Max>(&max_rank, sizeof(int));
+  rabit::Allreduce<op::Max>(&max_rank, 1);
```
trivialfis: Why is this 1?

src/allreduce_robust.h
```diff
@@ -286,12 +284,9 @@ class AllreduceRobust : public AllreduceBase {
         int role_flag = src[i].flag(SeqType::kCache) & dst[i].flag(SeqType::kCache);
         // if seqno is different in src and destination
         int seq_diff_flag = src[i].seqno() != dst[i].seqno() ? kDiffSeq : 0;
-        // if cache seqno is different in src and destination
-        int cache_diff_flag =
```
trivialfis: So this is no longer needed?

CodingCat:
oh @chenqin,  @trivialfis has a few questions here

test/model_recover.cc
chenqin: this is count, so we are syncing one int

chenqin:
> oh @chenqin, @trivialfis has a few questions here

working on a post follow up pr

src/allreduce_robust.h
chenqin: No, we don't need this anymore, will remove. 
Because we now don't allow write to cache after loadcheckpoint called after we remove is_bootstrap parameter from last pr

- If worker exit before loadcheckpoint, it start from beginning and get full copy of cache when it calls allreduce/broadcast; 
- if worker exit after loadcheckpoint and before first allreduce, it should functions exactly same as previous case; 
- if worker exit after loadcheckpoint and after first allreduce/broadcat, allreduce consensus protocol grantee the diff_seq() will be set & since recover worker doesn't finish loadcheckpoint yet, we know this is cache recover; if loadcheckpoint called, we no longer call load_cache hence not hitting this code.

trivialfis: Thanks for the detailed explanation.

PR #113, opened 2019-10-08
Expose RabitAllGatherRing and RabitGetRingPrevRank

nateagr:
Allreduce is great for dense tensors but can be highly inefficient for sparse tensors. Working with very big sparse tensors, I'd like to expose function allgatherring of allreduce_base.cc.

### DIFF
--- a/include/rabit/c_api.h
+++ b/include/rabit/c_api.h
@@ -44,6 +44,12 @@ RABIT_DLL bool RabitInit(int argc, char *argv[]);
 RABIT_DLL bool RabitFinalize(void);
 
 /*!
+ * \brief get rank of previous process in ring topology
+ * \return rank number of worker
+ * */
+RABIT_DLL int RabitGetRingPrevRank(void);
+
+/*!
  * \brief get rank of current process
  * \return rank number of worker
  * */
@@ -87,6 +93,30 @@ RABIT_DLL void RabitGetProcessorName(char *out_name,
  */
 RABIT_DLL void RabitBroadcast(void *sendrecv_data,
                               rbt_ulong size, int root);
+
+/*!
+ * \brief Allgather function, each node have a segment of data in the ring of sendrecvbuf,
+ *  the data provided by current node k is [slice_begin, slice_end),
+ *  the next node's segment must start with slice_end
+ *  after the call of Allgather, sendrecvbuf_ contains all the contents including all segments
+ *  use a ring based algorithm
+ *
+ * \param sendrecvbuf buffer for both sending and receiving data, it is a ring conceptually
+ * \param total_size total size of data to be gathered
+ * \param beginIndex beginning of the current slice in sendrecvbuf of type enum_dtype
+ * \param size_node_slice size of the current node slice
+ * \param size_prev_slice size of the previous slice i.e. slice of node (rank - 1) % world_size
+ * \param enum_dtype the enumeration of data type, see rabit::engine::mpi::DataType in engine.h of rabit include
+ * \return this function can return kSuccess, kSockError, kGetExcept, see ReturnType for details
+ * \sa ReturnType
+ */
+RABIT_DLL void RabitAllgather(void *sendrecvbuf,
+                                  size_t total_size,
+                                  size_t beginIndex,
+                                  size_t size_node_slice,
+                                  size_t size_prev_slice,
+                                  int enum_dtype);
+
 /*!
  * \brief perform in-place allreduce, on sendrecvbuf
  *        this function is NOT thread-safe
--- a/src/allreduce_base.h
+++ b/src/allreduce_base.h
@@ -15,9 +15,15 @@
 #include <vector>
 #include <string>
 #include <algorithm>
-#include "../include/rabit/internal/utils.h"
-#include "../include/rabit/internal/engine.h"
-#include "./socket.h"
+#include "rabit/internal/utils.h"
+#include "rabit/internal/engine.h"
+#include "rabit/internal/socket.h"
+
+#ifdef RABIT_CXXTESTDEFS_H
+#define private   public
+#define protected public
+#endif  // RABIT_CXXTESTDEFS_H
+
 
 namespace MPI {
 // MPI data type to be compatible with existing MPI interface
--- a/src/allreduce_base.h
+++ b/src/allreduce_base.h
@@ -529,7 +529,7 @@ class AllreduceBase : public IEngine {
   // version number of model
   int version_number;
   // whether the job is running in hadoop
-  int hadoop_mode;
+  bool hadoop_mode;
   //---- local data related to link ----
   // index of parent link, can be -1, meaning this is root of the tree
   int parent_index;
@@ -576,9 +576,13 @@ class AllreduceBase : public IEngine {
   // backdoor port
   int port = 0;
   // enable bootstrap cache 0 false 1 true
-  int rabit_bootstrap_cache = 0;
+  bool rabit_bootstrap_cache = false;
   // enable detailed logging
-  int rabit_debug = 0;
+  bool rabit_debug = false;
+  // by default, if rabit worker not recover in half an hour exit
+  int timeout_sec = 1800;
+  // flag to enable rabit_timeout
+  bool rabit_timeout = false;
   // Enable TCP node delay
   bool rabit_enable_tcp_no_delay = false;
 };
--- a/src/allreduce_base.h
+++ b/src/allreduce_base.h
@@ -579,6 +579,8 @@ class AllreduceBase : public IEngine {
   int rabit_bootstrap_cache = 0;
   // enable detailed logging
   int rabit_debug = 0;
+  // Enable TCP node delay
+  bool rabit_enable_tcp_no_delay = false;
 };
 }  // namespace engine
 }  // namespace rabit
--- a/src/allreduce_base.h
+++ b/src/allreduce_base.h
@@ -61,6 +61,10 @@ class AllreduceBase : public IEngine {
    */
   virtual void TrackerPrint(const std::string &msg);
 
+  /*! \brief get rank of previous node in ring topology*/
+  virtual int GetRingPrevRank(void) const {
+    return ring_prev->rank;
+  }
   /*! \brief get rank */
   virtual int GetRank(void) const {
     return rank;
@@ -77,6 +81,35 @@ class AllreduceBase : public IEngine {
   /*! \brief get rank */
   virtual std::string GetHost(void) const {
     return host_uri;
+  }
+
+  /*!
+  * \brief internal Allgather function, each node have a segment of data in the ring of sendrecvbuf,
+  *  the data provided by current node k is [slice_begin, slice_end),
+  *  the next node's segment must start with slice_end
+  *  after the call of Allgather, sendrecvbuf_ contains all the contents including all segments
+  *  use a ring based algorithm
+  *
+  * \param sendrecvbuf_ buffer for both sending and receiving data, it is a ring conceptually
+  * \param total_size total size of data to be gathered
+  * \param slice_begin beginning of the current slice
+  * \param slice_end end of the current slice
+  * \param size_prev_slice size of the previous slice i.e. slice of node (rank - 1) % world_size
+  * \param _file caller file name used to generate unique cache key
+  * \param _line caller line number used to generate unique cache key
+  * \param _caller caller function name used to generate unique cache key
+  */
+  virtual void Allgather(void *sendrecvbuf_, size_t total_size,
+                             size_t slice_begin,
+                             size_t slice_end,
+                             size_t size_prev_slice,
+                             const char* _file = _FILE,
+                             const int _line = _LINE,
+                             const char* _caller = _CALLER) {
+    if (world_size == 1 || world_size == -1) return;
+    utils::Assert(TryAllgatherRing(sendrecvbuf_, total_size,
+                                   slice_begin, slice_end, size_prev_slice) == kSuccess,
+                  "AllgatherRing failed");
   }
   /*!
    * \brief perform in-place allreduce, on sendrecvbuf
--- a/src/c_api.cc
+++ b/src/c_api.cc
@@ -13,7 +13,7 @@ namespace c_api {
 // helper use to avoid BitOR operator
 template<typename OP, typename DType>
 struct FHelper {
-  inline static void
+  static void
   Allreduce(DType *senrecvbuf_,
             size_t count,
             void (*prepare_fun)(void *arg),
@@ -25,7 +25,7 @@ struct FHelper {
 
 template<typename DType>
 struct FHelper<op::BitOR, DType> {
-  inline static void
+  static void
   Allreduce(DType *senrecvbuf_,
             size_t count,
             void (*prepare_fun)(void *arg),
@@ -35,11 +35,11 @@ struct FHelper<op::BitOR, DType> {
 };
 
 template<typename OP>
-inline void Allreduce_(void *sendrecvbuf_,
-                       size_t count,
-                       engine::mpi::DataType enum_dtype,
-                       void (*prepare_fun)(void *arg),
-                       void *prepare_arg) {
+void Allreduce_(void *sendrecvbuf_,
+                size_t count,
+                engine::mpi::DataType enum_dtype,
+                void (*prepare_fun)(void *arg),
+                void *prepare_arg) {
   using namespace engine::mpi;
   switch (enum_dtype) {
     case kChar:
@@ -85,12 +85,12 @@ inline void Allreduce_(void *sendrecvbuf_,
     default: utils::Error("unknown data_type");
   }
 }
-inline void Allreduce(void *sendrecvbuf,
-                      size_t count,
-                      engine::mpi::DataType enum_dtype,
-                      engine::mpi::OpType enum_op,
-                      void (*prepare_fun)(void *arg),
-                      void *prepare_arg) {
+void Allreduce(void *sendrecvbuf,
+               size_t count,
+               engine::mpi::DataType enum_dtype,
+               engine::mpi::OpType enum_op,
+               void (*prepare_fun)(void *arg),
+               void *prepare_arg) {
   using namespace engine::mpi;
   switch (enum_op) {
     case kMax:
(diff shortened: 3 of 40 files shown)


hcho3:
@nateagr Very interesting. I didn't realize that Rabit already had `AllreduceBase::TryAllgatherRing()`. I suppose it works like [MPI_Allgatherv](https://docs.microsoft.com/en-us/message-passing-interface/mpi-allgatherv-function) ? 

@thvasilo It looks like we have a basic implementation of Allgatherv in Rabit already. It doesn't appear to be fault-tolerant, however.

hcho3:
@nateagr Not sure how using Allgather directly would be more efficient, since you still need to set a fixed buffer size, so you'd end up over-allocating to fit the largest segment. Won't we need something like [MPI_Allgatherv](https://docs.microsoft.com/en-us/message-passing-interface/mpi-allgatherv-function) ?

nateagr:
> @nateagr Not sure how using Allgather directly would be more efficient, since you still need to set a fixed buffer size, so you'd end up over-allocating to fit the largest segment. Won't we need something like [MPI_Allgatherv](https://docs.microsoft.com/en-us/message-passing-interface/mpi-allgatherv-function) ?

When working with sparse tensors, non zero elements are most often at different indices for all nodes. For instance, let's consider an array of size N and a topology composed of 3 nodes. Node A could have 3 non zeros elements at index O, P and Q, node B could have 4 non zero elements at index R, S, T and U and node C could have only 1 non zero elements at index V. If N is relatively small, we could apply allreduce on the sparse vectors of nodes A, B and C without suffering any performance issue. But if N is really big, we would suffer performance issues regarding the small amount of non zero elements to reduce. In this last case, we'd rather perform allgather on non null elements and reduce locally. I hope this explanation makes sense. Otherwise, I could give further explanations

hcho3:
@nateagr Ah, got it. I understand that Allgather is definitely better than Allreduce, as communication amount goes down from `[matrix dimension]*[number of node]` to `max([number of nonzero elements in the i-th node])*[number of node]`. (Correct me if I'm wrong) So I will approve this PR.

However, I think you'd still have problem when the number of nonzero differs wildly between nodes. Let's say the matrix in Node A has 100 nonzero elements and Node B has 10. Then with Allgather, you'd need to allocate 100*2 elements in the recv buffer. Allgatherv (note the `v` suffix) allows for variable buffer sizes, so it would only allocate 100+10 elements. Unfortunately, Rabit doesn't yet have Allgatherv implementation, but it would be a great addition.

chenqin:
Looks good to me, please make sure surface on log and state using allgether is not fault tolerant.

thvasilo:
That's great. Although I don't know the use case you guys are talking
about, yes allgatherv is the preferred solution when nodes have wildly
different numbers of elements. We could hopefully use this as a base for an
allgatherv implementation.

--

hcho3:
@chenqin So only code in `allreduce_robust.cc` is fault tolerant?

chenqin:
> @chenqin So only code in `allreduce_robust.cc` is fault tolerant?

That's right. Only _robust has implementation of allreduce consensus protocol and checkpoint/restore implementation. As you might imagine, it is based on allreduce data exchange pattern.

src/allreduce_base.h
```diff
@@ -55,6 +55,10 @@ class AllreduceBase : public IEngine {
    */
   virtual void TrackerPrint(const std::string &msg);
 
+  /*! \brief get rank of previous node in ring topology*/
```
chenqin: please add those methods to _robust and _mock as well

nateagr:
> @nateagr Ah, got it. I understand that Allgather is definitely better than Allreduce, as communication amount goes down from `[matrix dimension]*[number of node]` to `max([number of nonzero elements in the i-th node])*[number of node]`. (Correct me if I'm wrong) So I will approve this PR.
> 
> However, I think you'd still have problem when the number of nonzero differs wildly between nodes. Let's say the matrix in Node A has 100 nonzero elements and Node B has 10. Then with Allgather, you'd need to allocate 100*2 elements in the recv buffer. Allgatherv (note the `v` suffix) allows for variable buffer sizes, so it would only allocate 100+10 elements. Unfortunately, Rabit doesn't yet have Allgatherv implementation, but it would be a great addition.

@hcho3: Implementation of AllgatherRing in Rabit works with variable buffer sizes. Following you examples, Node A would send 100 elements to Node B, Node B would send 10 elements to Node A and both will then reduce locally.

> please make sure surface on log and state using allgether is not fault tolerant.
@chenqin: What do you mean ?

nateagr:
Does someone understand the merge conflicts above ?

chenqin:
> > @nateagr Ah, got it. I understand that Allgather is definitely better than Allreduce, as communication amount goes down from `[matrix dimension]*[number of node]` to `max([number of nonzero elements in the i-th node])*[number of node]`. (Correct me if I'm wrong) So I will approve this PR.
> > However, I think you'd still have problem when the number of nonzero differs wildly between nodes. Let's say the matrix in Node A has 100 nonzero elements and Node B has 10. Then with Allgather, you'd need to allocate 100*2 elements in the recv buffer. Allgatherv (note the `v` suffix) allows for variable buffer sizes, so it would only allocate 100+10 elements. Unfortunately, Rabit doesn't yet have Allgatherv implementation, but it would be a great addition.
> 
> @hcho3: Implementation of AllgatherRing in Rabit works with variable buffer sizes. Following you examples, Node A would send 100 elements to Node B, Node B would send 10 elements to Node A and both will then reduce locally.
> 
> > please make sure surface on log and state using allgether is not fault tolerant.
> 
> @chenqin: What do you mean ?

As we can see the test is breaking because we need to implemented those methods in allreduce_robust and allreduce_mock as well.  When we add those new methods and call rabit from other project like xgboost, we need to notice user calling those methods will not have same level of fault tolerant guarantee as allreduce and broadcast enjoy.

nateagr:
Tests succeed on linux but they fail on macOS with the following message:
"No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.
2221Check the details on how to adjust your build configuration on: https://docs.travis-ci.com/user/common-build-problems/#Build-times-out-because-no-output-was-received"

Can I have some help ?

nateagr:
Hello! Could you please review ?

thvasilo:
If we're exposing new APIs we should have corresponding tests as in [allreduce](https://github.com/dmlc/rabit/tree/master/test/cpp) no?

trivialfis:
Will review soon.  Thanks for the patience.  First glance looks nice.

src/c_api.cc
```diff
@@ -120,8 +120,45 @@ inline void Allreduce(void *sendrecvbuf,
     default: utils::Error("unknown enum_op");
   }
 }
-
-
+inline void AllgatherRing(void *sendrecvbuf_,
```
trivialfis: Why do you need inline for this?  It might be some legacy convention floating around with it, but those are not necessary.

src/c_api.cc
```diff
@@ -120,8 +120,45 @@ inline void Allreduce(void *sendrecvbuf,
...
+                          size_t total_size,
+                          size_t beginIndex,
+                          size_t size_node_slice,
+                          size_t size_prev_slice,
+                          int enum_dtype) {
+  using namespace engine::mpi;
```
trivialfis: See above comment, can we move the definition into c header?

include/rabit/c_api.h
```diff
@@ -87,6 +93,30 @@ RABIT_DLL void RabitGetProcessorName(char *out_name,
...
+ * \param enum_dtype the enumeration of data type, see rabit::engine::mpi::DataType in engine.h of rabit include
+ * \return this function can return kSuccess, kSockError, kGetExcept, see ReturnType for details
+ * \sa ReturnType
+ */
+RABIT_DLL void RabitAllgatherRing(void *sendrecvbuf,
+                                  size_t total_size,
+                                  size_t beginIndex,
+                                  size_t size_node_slice,
+                                  size_t size_prev_slice,
+                                  int enum_dtype);
```
trivialfis: Can we move datatype definition into c_api? It's a little bit weird to ask others to look at a c++ header while using c api.

test/model_recover.cc
```diff
@@ -26,7 +26,7 @@ class Model : public rabit::Serializable {
   }
 };
 
-inline void TestMax(Model *model, int ntrial, int iter) {
+inline void TestMax(Model *model, int iter) {
```
chenqin: can we also add integration tests to 
https://github.com/dmlc/rabit/blob/master/test/test.mk#L12

include/rabit/c_api.h
```diff
@@ -87,6 +93,30 @@ RABIT_DLL void RabitGetProcessorName(char *out_name,
...
+ * \param sendrecvbuf buffer for both sending and receiving data, it is a ring conceptually
+ * \param total_size total size of data to be gathered
+ * \param beginIndex beginning of the current slice in sendrecvbuf of type enum_dtype
+ * \param size_node_slice size of the current node slice
+ * \param size_prev_slice size of the previous slice i.e. slice of node (rank - 1) % world_size
+ * \param enum_dtype the enumeration of data type, see rabit::engine::mpi::DataType in engine.h of rabit include
+ * \return this function can return kSuccess, kSockError, kGetExcept, see ReturnType for details
+ * \sa ReturnType
+ */
+RABIT_DLL void RabitAllgatherRing(void *sendrecvbuf,
```
chenqin: please use RabitAllgather(...) we might consider implement tree based allgather in future

src/c_api.cc
nateagr: Honestly, I've chosen to define this function as inline because similar functions but for allreduce are defined as inline. This function is very simple (computationally) and has only one caller. Defining this function as inline won't do any harm. It would avoid context switching between the caller and this function if the compiler agrees to inline this function. But if you prefer, I can define it as a regular function.

trivialfis: @nateagr  inline is a specifier for linker not for compiler despite its name.  See https://stackoverflow.com/questions/1759300/when-should-i-write-the-keyword-inline-for-a-function-method  It doesn't do any harm but just not necessary in 2019.  I won't block the PR for this small detail as you put a lots of effort and It's really useful changes.  But it's also not harmful to clean up the code a little bit. ;-)

src/c_api.cc
nateagr: Done

test/model_recover.cc
nateagr: What integration tests do you think should be added ?

src/c_api.cc
nateagr: Ok I'll do the cleanup in c_api.cc then. Thanks for the clarification.

nateagr:
Hello! I'd like to give you more context regarding the last commit "Enabled TCP_NODELAY to decrease latency". When benchmarking rabit ops (allreduce and allgather) with different inputs (different sizes), I observed that the latency sometimes increased from 1-2 ms to 40 ms for small packets. After a bit of research, I figured this was due to one of Nagle's TCP algorithms. You can find more details here: https://www.extrahop.com/company/blog/2016/tcp-nodelay-nagle-quickack-best-practices/

thvasilo:
Hello @nateagr, one design question I had is why it's necessary to have access to the size of the data of the previous node in the ring.

I understand how that makes a ring-based allgather easier to implement, but other implementations of allgather that I've seen (for example in MPI and IIRC in LightGBM) do not have that requirement.

You mentioned that you're working on a tree-based allgather as well, will that remove this requirement and have an API closer to MPI_allgather?

Also could you provide an end to end example of using this version of allgather for aggregating sparse arrays from multiple nodes?

My understanding currently is something like (pseudocode):

```
// Create an array to be populated with the number of elements from each rank
int num_elements[WORLD_SIZE]

// Change one element in the array with my (local) number of elements
num_elements[MY_RANK] = local_num_elements

// Allreduce/sum this array to get the number of elements from each worker
allreduce<Sum>(num_elements)
// Question: Can the above be skipped given that we have access to the rank of the previous worker in the ring?

// Get the number of elements in the previous worker
int prev_elements = num_elements[rabit:::RabitGetRingPrevRank()]
int total_elements = std::accumulate(num_elements)
int offset = std::partial_sum(num_elements[0], num_elements[MY_RANK])

// Now we have all the information to do a ring sparse allgather
int all_elements[total_elements]
RabitAllgather(all_elements, total_elements, offset, local_num_elements, prev_elements, rabit::kInt)
```

Is my pseudocode correct?

test/model_recover.cc
chenqin: we should cover Allgather operation with mock=0,<op of allgather>,.. and see if recovery works properly.

nateagr:
> Hello @nateagr, one design question I had is why it's necessary to have access to the size of the data of the previous node in the ring.
> 
> I understand how that makes a ring-based allgather easier to implement, but other implementations of allgather that I've seen (for example in MPI and IIRC in LightGBM) do not have that requirement.
> 
> You mentioned that you're working on a tree-based allgather as well, will that remove this requirement and have an API closer to MPI_allgather?
> 
> Also could you provide an end to end example of using this version of allgather for aggregating sparse arrays from multiple nodes?
> 
> My understanding currently is something like (pseudocode):
> 
> ```
> // Create an array to be populated with the number of elements from each rank
> int num_elements[WORLD_SIZE]
> 
> // Change one element in the array with my (local) number of elements, the rest should be 0
> num_elements[MY_RANK] = local_num_elements
> 
> // Allreduce/sum this array to get the number of elements from each worker
> allreduce<Sum>(num_elements)
> // Question: Can the above be skipped given that we have access to the rank of the previous worker in the ring?
> 
> // Get the number of elements in the previous worker
> int prev_elements = num_elements[rabit::RabitGetRingPrevRank()]
> // Get the total number of elements
> int total_elements = std::accumulate(num_elements)
> // Get the offset, i.e. the start point for the local worker's data
> int offset;
> std::partial_sum(num_elements[0], num_elements[MY_RANK], offset)
> 
> // Now we have all the information to do a ring sparse allgather
> int all_elements[total_elements]
> RabitAllgather(all_elements, total_elements, offset, local_num_elements, prev_elements, rabit::kInt)
> ```
> 
> Is my pseudocode correct?

Hello @thvasilo. Your pseudocode is almost correct:
- Number of elements to gather from other nodes are retrieved by calling allgather instead of allreduce. The number of elements of the previous node in the ring for this first call to allgather is implicit, it is 1.
- all_elements must be populated with local node data (starting from position offset to offset + local_num_elements) before calling allgather for the second time.
To get rid of the parameter prev_elements in the public api, we could retrieved the number of elements to gather from other nodes by calling allgather internally. It makes the public api easier so I'll do it.

thvasilo:
Yeah, for example simplifying such calls is the purpose of the [mxx library](https://github.com/patflick/mxx) (for MPI), I can recommend taking a look at it for ideas. [One of their implementations of allgatherv](https://github.com/patflick/mxx/blob/14b4f0dda21c94a9a9083d6adeed39023c08f3d1/include/mxx/collective.hpp#L1121) for example is super intuitive, you just provide an outgoing vector and the implementation takes care of the rest (gathering number of elements from other processes and returning the result vector with all values).

There are tradeoffs of course (that call duplicates memory by returning a new object) but it's good to have options.

> all_elements must be populated with local node data (starting from position offset to offset + local_num_elements) before calling allgather for the second time.

This would also duplicate memory no? Unless we explicitly clear the local buffer after/during the aggregation. I guess the only way to avoid that would be using something like the MPI [in-place operator](https://www.open-mpi.org/doc/v3.0/man3/MPI_Allgather.3.php#sect8) which has other limitations.

thvasilo:
That's up to the maintainers to decide in the end. If I find the time I'll
try to run some tests on an MPI cluster, can't promise to do it on YARN too
though, but I feel like somebody should test that too.

Also, because we are working on something similar in a research capacity
(working on a related paper) would you mind sending me an email at
tvas@kth.se?
I'll fill you in on the details there.

Thanks,
Theo

trivialfis:
Will take a look tomorrow.  Thanks :-)

thvasilo:
Small thing I just noticed the `.cpp` test files in this PR are duplicating existing `.cc` files. All other files in the project use `.cc` so we should be consistent.

I'm gonna try running some tests on an MPI cluster today, will report back.

test/cpp/allreduce_mock_test.cpp
```diff
@@ -0,0 +1,51 @@
...
+  std::string mock_str = "mock=3,13,22,0";
+  char cmd[mock_str.size()+1];
+  std::copy(mock_str.begin(), mock_str.end(), cmd);
+  cmd[mock_str.size()] = '\0';
+  char* argv[] = {cmd};
+  m.Init(1, argv);
+  m.rank = 3;
+  m.version_number=13;
+  m.seq_counter=22;
+  EXPECT_EXIT(m.AllgatherRing(nullptr,0,0,0,0), ::testing::ExitedWithCode(255), "");
```
thvasilo: `AllGatherRing` is not part of the API, this will not compile.

```suggestion
  EXPECT_EXIT(m.Allgather(nullptr,0,0,0,0), ::testing::ExitedWithCode(255), "");
```

thvasilo:
I ran the provided integration test today (`model_recover`) on an MPI cluster with 10M elements using 

```
../dmlc-core/tracker/dmlc-submit --cluster slurm --num-workers 3  --host-file /shared/hostnames.txt model_recover 10000000 rabit_bootstrap_cache=true rabit_debug=true
```

and all the tests seem to pass. 

Trying to use 100M elements fails with memory errors for allgather, because we try to gather `num_elements * num_proc * iter` floats. This differentiates the allgather test from the other tests (e.g. TestSum, TestMax) that work with `num_elements * iter` elements.
Perhaps after merging we want to change the allgather test to use as many elements as the other tests by gathering only parts of the existing model, because right now it will run into OOM errors before all others do.

Another test I'd like to include perhaps after merging this is a test with uneven arrays, where we gather different numbers of elements from each worker as described above.

Apart from the mock tests I haven't tried recovery and I'm not sure if that's even possible using the SLURM tracker.

chenqin:
> I ran the provided integration test today (`model_recover`) on an MPI cluster with 10M elements using
> 
> ```
> ../dmlc-core/tracker/dmlc-submit --cluster slurm --num-workers 3  --host-file /shared/hostnames.txt model_recover 10000000 rabit_bootstrap_cache=true rabit_debug=true
> ```
> 
> and all the tests seem to pass.
> 
> Trying to use 100M elements fails with memory errors for allgather, because we try to gather `num_elements * num_proc * iter` floats. This differentiates the allgather test from the other tests (e.g. TestSum, TestMax) that work with `num_elements * iter` elements.
> Perhaps after merging we want to change the allgather test to use as many elements as the other tests by gathering only parts of the existing model, because right now it will run into OOM errors before all others do.

should we try use rabit_reduce_buffer=4GB in command line?
https://github.com/dmlc/rabit/blob/master/src/allreduce_base.cc#L40

thvasilo:
> should we try use rabit_reduce_buffer=4GB in command line?
https://github.com/dmlc/rabit/blob/master/src/allreduce_base.cc#L40

What is the purpose of this buffer? I should not that my comment above I consider to be "expected behavior", I was just noting that because of the way the test is set up, the other tests can complete but the allgather test consumes `num_workers` more memory.

Is this exposed to the user in some way or is it hardcoded?

test/cpp/allreduce_mock_test.cpp
nateagr: Right. Thanks.

nateagr:
> I ran the provided integration test today (`model_recover`) on an MPI cluster with 10M elements using
> 
> ```
> ../dmlc-core/tracker/dmlc-submit --cluster slurm --num-workers 3  --host-file /shared/hostnames.txt model_recover 10000000 rabit_bootstrap_cache=true rabit_debug=true
> ```
> 
> and all the tests seem to pass.
> 
> Trying to use 100M elements fails with memory errors for allgather, because we try to gather `num_elements * num_proc * iter` floats. This differentiates the allgather test from the other tests (e.g. TestSum, TestMax) that work with `num_elements * iter` elements.
> Perhaps after merging we want to change the allgather test to use as many elements as the other tests by gathering only parts of the existing model, because right now it will run into OOM errors before all others do.
> 
> Another test I'd like to include perhaps after merging this is a test with uneven arrays, where we gather different numbers of elements from each worker as described above.
> 
> Apart from the mock tests I haven't tried recovery and I'm not sure if that's even possible using the SLURM tracker.

Hi there. After this pull request, I'll modify the test as you said to not run into OOM. Also, I'll add another test with uneven arrays to gather. I can't do it right now unfortunately because I have limited access to internet right now for the next 2 weeks.

Please let me know if there is  anything that might block this pull request to be merged.

trivialfis:
@chenqin @hcho3 Any update on this?  :-)

nateagr:
Hi there. Can I have an update on this PR please.

trivialfis:
Thanks!

PR #119, opened 2019-10-15
Fix compilation failure on Windows

hcho3:
`strcasecmp` is POSIX so does not exist on Windows. Use `_stricmp` instead. This is the cause of the compilation failure in https://xgboost-ci.net/blue/organizations/jenkins/xgboost-win64/detail/PR-4876/50/pipeline#step-44-log-401

@trivialfis @chenqin

### DIFF
--- a/include/rabit/internal/utils.h
+++ b/include/rabit/internal/utils.h
@@ -66,7 +66,7 @@ const int kPrintBuffer = 1 << 12;
 /*! \brief we may want to keep the process alive when there are multiple workers
  * co-locate in the same process */
 extern bool STOP_PROCESS_ON_ERROR;
-  
+
 /* \brief Case-insensitive string comparison */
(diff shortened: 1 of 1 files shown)


include/rabit/internal/utils.h
```diff
@@ -66,10 +66,19 @@ const int kPrintBuffer = 1 << 12;
 /*! \brief we may want to keep the process alive when there are multiple workers
  * co-locate in the same process */
 extern bool STOP_PROCESS_ON_ERROR;
+
```
chenqin: include/rabit/internal/utils.h:69:  Line ends in whitespace.  Consider deleting these extra spaces.  [whitespace/end_of_line] [4]

chenqin:
looks good overall, please fix lint error

PR #120, opened 2019-10-15
throw dmlc::Error 

chenqin:
XGBoost JNI only handles dmlc::Error from try catch clause, replace std::runtime_error with dmlc::Error could allow us gracefully handle rabit worker assertion errors and propagated to jvm layer

https://github.com/dmlc/xgboost/blob/master/src/c_api/c_api_error.h#L17
image

image

@CodingCat @trivialfis @hcho3


chenqin:
@hcho3 @trivialfis this should be straight forward change, can you guys help review?

chenqin:
Thanks a lot!

PR #122, opened 2019-10-21
simplify parameter timeout and ring_reduce

chenqin:
per feedback from https://github.com/dmlc/xgboost/pull/4966 
- simplify rabit_timeout, opt-in if user set a positive value
- simplify opt-in ring based communication setting with bool


chenqin:
@CodingCat @trivialfis straightforward one

chenqin:
per convo, pull back due to behavior change concerns.

PR #124, opened 2019-10-21
Fix gtest include

thvasilo:
The unit tests currently do not include the Gtest headers leading to compilation failure.


hcho3:

PR #125, opened 2019-10-21
allow timeout to 0 for immediate exit

chenqin:
@CodingCat

### DIFF
--- a/src/allreduce_base.cc
+++ b/src/allreduce_base.cc
@@ -219,7 +219,7 @@ void AllreduceBase::SetParam(const char *name, const char *val) {
   }
   if (!strcmp(name, "rabit_timeout_sec")) {
     timeout_sec = atoi(val);
-    utils::Assert(timeout_sec > 0, "rabit_timeout_sec should be greater than 0 second");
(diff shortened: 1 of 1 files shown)


chenqin:
@trivialfis @hcho3 oneliner

CodingCat:
what does  immediate exit  mean here?

chenqin:
when rabit worker observed peers socket error, it starts countdown and exit if failed worker fail to connect before timeout expire. This is causing issues with xgb4j-spark jobs where we observed worker see retry timeout because we close sparkcontext without before rabit agent and agent keep ping already closed tracker. 
https://github.com/dmlc/rabit/blob/master/src/allreduce_robust.cc#L641

we when set timeout=0 all worker will failure at same time if a worker dead. It also helps with testing in xgb4j-spark side and simulate graceful failure handling.

CodingCat:
> close sparkcontext before rabit agent and agent keep ping already closed tracker. 

when we close SparkContext, executors will quit, therefore the tasks in executor side will also quit, and rabit worker is running in a thread of executors, why it can keep pinging tracker?

CodingCat:
per offline, LGTM

PR #128, opened 2019-11-12
allow duplicated bootstrap allreduce overwrite previous results

chenqin:
Rabit assume bootstrap allreduce has unique signature, in https://github.com/dmlc/xgboost/issues/5012 we find it failed to satisfy such constraint. 

The goal of this pr is to loosen such constraint and allow last write overwrite results with same signature during bootstrap phase. Here is example https://github.com/dmlc/xgboost/pull/4990/files#diff-5fabd019d3f1af7572baa4a6301cf076R36

### DIFF
--- a/src/allreduce_robust.cc
+++ b/src/allreduce_robust.cc
@@ -108,7 +108,9 @@ int AllreduceRobust::SetBootstrapCache(const std::string &key, const void *buf,
       break;
     }
   }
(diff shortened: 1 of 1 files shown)


chenqin:
@trivialfis

chenqin:
> Thanks. Could you create a new issue in XGBoost?

https://github.com/dmlc/xgboost/issues/5033

PR #129, opened 2019-11-25
fix xgboost build failure introduced by allgather interface

chenqin:
detail change https://github.com/dmlc/rabit/pull/113

### DIFF
--- a/include/rabit/internal/rabit-inl.h
+++ b/include/rabit/internal/rabit-inl.h
@@ -207,7 +207,7 @@ inline void Allgather(DType *sendrecvbuf,
                       const char* _file,
                       const int _line,
                       const char* _caller) {
-  engine::Allgather(sendrecvbuf, totalSize * sizeof(DType), beginIndex * sizeof(DType),
+  engine::GetEngine()->Allgather(sendrecvbuf, totalSize * sizeof(DType), beginIndex * sizeof(DType),
                         (beginIndex + sizeNodeSlice) * sizeof(DType),
                         sizePrevSlice * sizeof(DType), _file, _line, _caller);
 }
--- a/include/rabit/internal/rabit-inl.h
+++ b/include/rabit/internal/rabit-inl.h
@@ -196,17 +196,22 @@ inline void Allreduce(DType *sendrecvbuf, size_t count,
                      engine::mpi::GetType<DType>(), OP::kType, InvokeLambda_, &prepare_fun,
                      _file, _line, _caller);
 }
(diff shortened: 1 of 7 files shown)


chenqin:
@trivialfis

thvasilo:
Pinging @nateagr too. This changes seems strange to me, could you explain briefly why all gather needs these extra inputs to function?

chenqin:
> Pinging @nateagr too. This changes seems strange to me, could you explain briefly why all gather needs these extra inputs to function?

those extra inputs were used to ensure we can do single worker failure recovery before loadcheckpoint called.

include/rabit/internal/rabit-inl.h
```diff
@@ -196,17 +196,22 @@ inline void Allreduce(DType *sendrecvbuf, size_t count,
...
                      _file, _line, _caller);
 }
-#endif  // C++11
 
 // Performs inplace Allgather
 template<typename DType>
-inline void Allgather(DType *sendrecvbuf, size_t totalSize,
-                      size_t beginIndex, size_t sizeNodeSlice,
-                      size_t sizePrevSlice) {
-  engine::Allgather(sendrecvbuf, totalSize * sizeof(DType), beginIndex * sizeof(DType),
```
chenqin: this is rootcause, which point to undeclared function

src/c_api.cc
```diff
@@ -131,33 +131,54 @@ void Allgather(void *sendrecvbuf_,
...
     break;
   case kDouble:
     type_size = sizeof(double);
+    rabit::Allgather(static_cast<double*>(sendrecvbuf_), total_size * type_size,
+      beginIndex * type_size, (beginIndex + size_node_slice) * type_size,
+      size_prev_slice * type_size);
     break;
   default: utils::Error("unknown data_type");
   }
-  engine::Allgather
```
chenqin: same as above

src/engine_empty.cc
```diff
@@ -26,17 +26,18 @@ class EmptyEngine : public IEngine {
...
+                         size_t slice_end,
+                         size_t size_prev_slice,
+                         const char* _file,
+                         const int _line,
+                         const char* _caller) {
     utils::Error("EmptyEngine:: Allgather is not supported");
   }
   virtual int GetRingPrevRank(void) const {
     utils::Error("EmptyEngine:: GetRingPrevRank is not supported");
+    return -1;
```
chenqin: a type fix

PR #130, opened 2019-12-01
change include file path

CodingCat:
not sure why we had this change...but I found it broke the original command `mvn package` running in mac

@trivialfis @chenqin

### DIFF
--- a/include/rabit/internal/engine.h
+++ b/include/rabit/internal/engine.h
@@ -7,7 +7,7 @@
 #ifndef RABIT_INTERNAL_ENGINE_H_
 #define RABIT_INTERNAL_ENGINE_H_
 #include <string>
-#include "rabit/serializable.h"
+#include "../serializable.h"
 
 #if (defined(__GNUC__) && !defined(__clang__))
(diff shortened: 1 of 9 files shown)


chenqin:
LGTM, I thought we had rabit issue with cmake file.

trivialfis:
I'm not sure about the relative paths in public headers.  We have `-Idmlc-core/include`, some headers in there can sneak into the translation unit replacing the one in `rabit`. While using `rabit/serialisation.h` provide some weak guarantees that no such error can occur silently.

For the reason why this could happen, include path searching starts with current file, but not limited to it.  If the compiler fails to find one in relative path to current file, it will start again from paths specified by `-I` flag.  I would like to resolve your build issue first, it may indicates errors described above is actually happening in our code base.

trivialfis (commented):
Could you please provide the error message you encountered so I can take a look?

CodingCat:
> Could you please provide the error message you encountered so I can take a look?

it actually simply cannot find files like `rabit/serializable.h`

hcho3:

Issue #127, opened 2019-11-04
C vs C++ MPI Usage


wrathematics:
As you know, the MPI standard deprecated C++ bindings. Some implementations still offer them (e.g. OpenMPI) while others do not (e.g. MSMPI). Would you be receptive to a re-write of the MPI-backend internals to use the MPI C API? Looking at https://github.com/dmlc/rabit/issues/31 it seems like you're open to the change at least in principle. I am willing to do this, but I wanted to make sure before I started.

hcho3:
@trivialfis @chenqin

trivialfis:
Thanks!  PRs are welcomed!  Just currently due to MPI backend doesn't support fault tolerance so we don't really use it.  @chenqin may provide more inputs here.

chenqin:
> Thanks! PRs are welcomed! Just currently due to MPI backend doesn't support fault tolerance so we don't really use it. @chenqin may provide more inputs here.

Thanks Travis. MPI were not prioritized because it doesn’t support fault recovery we expected from socket implementation. I put some thoughts on building a overlay on top and founds it’s not very straightforward. Meanwhile, we are happy to work with you on this if that’s what’s your passioned about.

thvasilo:
If it's just a straightforward translation from the C++ calls to the C ones I don't see why we wouldn't do it. 

Note that AFAI understand, my research use cases still make use of the sockets version for communication, I'm only using MPI/SLURM as a tracker.

Re. what @chenqin suggested I've started recently looking into LightGBM and it has a nice thin layer between MPI and their sockets implementation, with collectives making heavy use of a sendrecv operation internally, meaning that it doesn't take too much duplication to support the two.

For the future, it might be worth taking a look.

wrathematics:
Great, thanks!

One thing I would add is that changing to manual send/recv patterns will likely come at the cost of performance. I think most MPI implementations use recursive doubling for allreduce, which is a communication-avoiding strategy. And on HPC systems with boutique interconnects, collectives are optimized even further to take advantage of the network topology.

thvasilo:
Agreed, no point in reinventing the wheel, we should keep it as high level as possible.

snoweye:
Please see this [PR](https://github.com/dmlc/rabit/pull/135) for the MPI C implementation and application.

hcho3:
[...]

PR #135, opened 2020-02-16
Mpicc

snoweye:
- MPI C implementations with an example tested with OpenMPI and CI in Linux and osx.
- The application can be found in [pbdXGB](https://github.com/snoweye/pbdXGB) with CI tested in Linux (OpenMPI) and Widnows (MS-MPI).


trivialfis:
Thanks for the PR and sorry for the wait!  Feel free to ping me in the future ;-).  Is it possible to completely replace the original c++ implementation?

snoweye:
Yes, the implementation is a complete replacement except some difference where
- `handle_` and `htype_` are altered from `pointer to class pointing to MPI struct` to `pointer pointing to MPI struct` if my understanding is correct, and
- `handle_` may not be automatically free (potential memory leak) as in C++ via `ReduceHandle::~ReduceHandle`.

I can get rid of C++ and simplify those unnecessary steps wherever is possible. However, this had not been discussed as in the issue [#127](https://github.com/dmlc/rabit/issues/127). Would others agree with the changes?

@trivialfis @hcho3 @chenqin @thvasilo @tqchen @wrathematics

trivialfis:
@snoweye I think there's an implementation with mpi backend here.  Can we just replace that without any architectural change?  Feel free to point out the issues here.

snoweye:
It is difficult without architectural change, though, possibly ends up rewrite/redefine most what MPI C++ deprecated binding did if keeping those MPI namespace is needed. This PR is to keep those MPI namespace when no mpi is needed, but it switches the binding where mpi or speed is needed.

snoweye:
I have the other branch [`mpi_replace`](https://github.com/snoweye/rabit/tree/mpi_replace) that replace the MPI C++ code. This should work though I did not test yet.

However, it is not completely without changing architectural because I have trouble to get ride of these lines regarding [`ReduceFunction`](https://github.com/snoweye/rabit/blob/mpi_replace/include/rabit/internal/engine.h#L51-L59). Do you have any idea to resolve or by pass it? Either redefine it or cast correctly to the [`*fp`](https://github.com/snoweye/rabit/blob/mpi_replace/src/engine_mpi.cc#L265) when the ReduceHandle is initialized.

@trivialfis

snoweye:
@trivialfis Would you take a quick look the new changes in the branch [`mpi_replace`](https://github.com/snoweye/rabit/tree/mpi_replace) and see if that is what you expect for or else? I will do the test if there is an ok to go for.

hcho3:
@trivialfis Can we merge this for now, given that this doesn't modify existing code but adds a new functionality? We can even mark it as "experimental". On a related note, we can consider adding MPI in the XGBoost testing pipeline.

trivialfis:
No.  I don't want to add new API to rabit.  Allow me to explain a little.  Right now I have a huge headache over rabit's model recovery functionality.  It saves model and model only, so anything in DMatrix is not check pointed.  That's why it doesn't work for dask because we requires DMatrix to handle itself.  Also there's an unclear behaviour regarding repeated calls before training.  Adding more APIs will only make things more difficult to work with regarding model recovery.

hcho3:

PR #139, opened 2020-05-07
Set a minimal reducer size and parent_down size

FelixYBW:
currently distributed xgboost's performance varies largely run to run even we tried to input exact the same partition into each worker. Its performance is more than 2x poor if we wrongly config to use 1GbE network (It's another issue we need to fix), or if we run on a IPoIB network.

The root cause is that sometimes a message is broken into two pieces when a child receives it, but the second piece has a long wait up to >10 ms, while the network is empty actually.
Looks like the poll(-1) caused this issue but I set it to poll(0)+pause which doesn't help.

image

This happens more frequently on 1GbE network and IPoIB, which leads to even poor performance.

The fix is to set a minimal reducer size as N Byte, so every time parent can reduce and pass down the N Byte message at least. Then child can wait until receiving N Byte. In our test, xgboos uses allreduce(sum), the real calculation time is very little so we can set it upto 1MB which is bigger than the maximal msage size of our workload.

image

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -63,7 +63,7 @@ addons:
 
 before_install:
   - git clone https://github.com/dmlc/dmlc-core
-  - export TRAVIS=dmlc-core/scripts/travis/
+  - export TRAVIS=./scripts/
   - source ${TRAVIS}/travis_setup_env.sh
   - ${TRAVIS}/travis_osx_install.sh
   - source ./scripts/travis_setup.sh
--- a/include/rabit/base.h
+++ b/include/rabit/base.h
@@ -1,0 +1,19 @@
+/*!
+ * Copyright (c) 2020 by Contributors
+ * \file base.h
+ * \brief Macros common to all headers
+ *
+ * \author Hyunsu Cho
+ */
+
+#ifndef RABIT_BASE_H_
+#define RABIT_BASE_H_
+
+#ifndef _CRT_SECURE_NO_WARNINGS
+#define _CRT_SECURE_NO_WARNINGS
+#endif  // _CRT_SECURE_NO_WARNINGS
+#ifndef _CRT_SECURE_NO_DEPRECATE
+#define _CRT_SECURE_NO_DEPRECATE
+#endif  // _CRT_SECURE_NO_DEPRECATE
+
+#endif  // RABIT_BASE_H_
--- a/include/rabit/c_api.h
+++ b/include/rabit/c_api.h
@@ -18,7 +18,7 @@
 #if defined(_MSC_VER) || defined(_WIN32)
 #define RABIT_DLL RABIT_EXTERN_C __declspec(dllexport)
 #else
-#define RABIT_DLL RABIT_EXTERN_C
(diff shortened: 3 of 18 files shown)


FelixYBW:
I did an experiment of epoll, but performance is the same even if I use edge_trigger. We still need the busy recv fix here. A epoll_wait + recv still leads to delays. 
Some questions are still open:
- It mainly happens on parent_down. Not frequently happen on receiving child_up.
- This fix solved most of the delay, but not all. Sometimes the whole message is delayed.  It may because the network hardware delay (less likely) or because OS scheduling.
- Why the busy poll (poll with 0 timeout) doesn't solve the problem.

trivialfis:
Let me fix the travis tests.

trivialfis:
CI should be fixed in https://github.com/dmlc/rabit/pull/140 .

FelixYBW:
what's the error? style check on my local machine passed.

src/engine_mpi.cc:13:  Found C++ system header after other system header. Should be: engine_mpi.h, c system, c++ system, other.  

The file isn't touched.

11 #define NOMINMAX
12 #include <mpi.h>
13 #include <cstdio>
14 #include "rabit/internal/engine.h"

hcho3:
Cpplint 1.5.0 is released last week and it changed the way it checks the header order.

hcho3:
CI is fixed in #141

FelixYBW:
Passed eventually. It's odd I have to move <mpi.h> before <cstdio>. Added <string>
fix allreduce_base.cc's header seq

FelixYBW:
@trivialfis @hcho3 

Any comments on the PR? Can we merge it?

hcho3:
@FelixYBW Was this change tested in a production cluster with XGBoost (or other application)?

FelixYBW:
> @FelixYBW Was this change tested in a production cluster with XGBoost (or other application)?

Yes, it's tested in several clusters with xgboost training.

FelixYBW:
> @FelixYBW Was this change tested in a production cluster with XGBoost (or other application)? I ask this question because the plot you posted shows only the time for sending messages. The CI set up here may not detect all possible ways this change can break XGBoost.
> 
> > In our test, xgboos uses allreduce(sum), the real calculation time is very little so we can set it upto 1MB which is bigger than the maximal msage size of our workload.
> 
> Can I take it as that you tested with XGBoost? How is the impact on the end-to-end training performance?

Yes, we tested on xgboost. It works well so far. The chart in the first post is the end2end xgboost training performance improvement on 3 clusters with 4 nodes in each cluster but using different network. The dataset is a 600M Row x 46 columns one. Also the patch makes xgboost training time much more stable.

hcho3:
> The chart in the first post is the end2end xgboost training performance improvement on 3 clusters with 4 nodes in each cluster but using different network.

@FelixYBW Nice, thanks for clarifying. 1.5-3x boost end-to-end is quite nice.

hcho3:
@FelixYBW Merged. Thanks! I'll file a pull request to use this commit in XGBoost.

PR #144, opened 2020-08-13
Use shared ptr for net link.

trivialfis:
I ran into an issue with expired pointer in `tree_links`.  So far the issue is not reproducible on JenkinsCI and I can't find the root cause.  So I changed everything to shared ptr ...


trivialfis:
@RAMitchell

RAMitchell (approved):
Looks okay in my non-expert opinion.

hcho3:

